PR #92, opened 2013-11-14
Use /:path instead of ?path=:path to get contents

raphink:
Passing `path` in the query string does not seem to work anymore with the Github API (I'm admittedly guilty of not seeking to know why). The [API documentation](http://developer.github.com/v3/repos/contents/) says to use:

    GET /repos/:owner/:repo/contents/:path

which is what this patch does (and it works ;-)).

PR #98, opened 2014-01-03
Update content method to use v3 spec. Fixes #92.

coreybutler:
Updates the repository content method according to the Github API v3 spec.

### DIFF
--- a/github.js
+++ b/github.js
@@ -440,7 +440,7 @@
       // --------
 
       this.contents = function(branch, path, cb, sync) {
-        return _request("GET", repoPath + "/contents?ref=" + branch + (path ? "&path=" + path : ""), null, cb, 'raw', sync);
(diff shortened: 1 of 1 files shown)


aendrew:
Does this still need to be merged? If so, I'd like to do so sometime soon. Thanks!

aendrew:
I'm pretty sure this is outdated and am thus closing. 

@w35l3y — blame says that your commit from PR #92 is what's currently in HEAD. Given the totally different syntax of this PR, my belief is that you adapted it into what was eventually merged. Can you confirm [this](https://github.com/michael/github/blame/master/github.js#L582) supersedes this PR? Thanks!

w35l3y:
All I remember is that I changed it a little. I didn't find it in the blame.

PR #101, opened 2014-01-16
Fix the contents function on the repository API.

darktears:
It appears that Github either changed their way of querying files
or directory or the bug was always there but say you want to do :

repo.contents(branch, "MYFILE", function(err, contents) {}, false);

it currently doesn't work and list only the root directory of the
repository. Changing slightly the way Github.js query the contents
make things works correctly as expected.

### DIFF
--- a/github.js
+++ b/github.js
@@ -440,7 +440,7 @@
       // --------
 
       this.contents = function(branch, path, cb, sync) {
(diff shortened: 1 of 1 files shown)


aendrew:
Both #92 and #127 solve this. If you can beat them to updating your PR, I'll merge this instead of those. :checkered_flag:

aendrew:
Closing as duplicate of merged PR.

PR #102, opened 2014-01-16
basic issues API

avyaznikov:


aendrew:
@w35l3y Wouldn't say so; the existing Issues API doesn't have anywhere this level of functionality, would be happy to merge it. That said, I definitely want some unit tests before doing so. Thanks!

@avyaznikov Mind creating a test case for your changes, and merging them into the existing Issues API? Thanks!

aendrew:
Actually, this is pretty depreciated. The only benefit is the `open` and `close`, which can probably be handled via `Issue.list`. Closing, feel free to comment if you feel it should be reopened and reworked. Thanks!

PR #113, opened 2014-03-30
content API update and some other features

whoisterencelee:
Instead of using read/write, the content API should be used unless the data is more than 1MB, because the number of http requests get reduced from 3 to just 1, and allows multi-user commits which tracks the document SHA (prevents update with a old SHA).

A few changes to allow 4 things:
1) fix for issue #101 
getting Content(branch, path, cb, sync, 'json') will get a json with file obj sha, which is closer to what http://developer.github.com/v3/repos/contents/#get-contents is describing, with option 'json' will give the same behaviour as before, getting the file raw content

2) added a repo.req function to allow access to request, which is useful if anyone wants to make custom API calls

3) allow request status==0 which is returned from XHR request made from a locally loaded page

4) getCommits should use pagination

### DIFF
--- a/github.js
+++ b/github.js
@@ -645,9 +645,10 @@
           }
           _request("GET", url, null, cb);
       };
-    };
 
-    this.req=_request;
+      this.req=_request;
+
+    };
 
     // Gists API
     // =======
--- a/github.js
+++ b/github.js
@@ -445,9 +445,17 @@
       // --------
 
       this.contents = function(branch, path, cb, sync, datatype) {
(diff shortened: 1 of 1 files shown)


aendrew:
@whoisterencelee Would you mind solving the merge conflicts and rebasing if you still would like this added? It seems like a worthwhile PR and I'd like to add it to 0.10.8 if possible. Apologies for the delay!

aendrew:
Actually, reviewing this a bit more, it's effectively duplicating existing PR. Closing as duplicate.

PR #114, opened 2014-04-02
add listStatuses and listHeads. re-implement listBranches using listHeads

afex:
listHeads allows access to the extra metadata returned from /git/refs/heads, whereas listBranches only returns branch names

listStatuses returns commit statuses for a given ref

### DIFF
--- a/github.js
+++ b/github.js
@@ -306,13 +306,33 @@
         });
       };
 
-      // List all branches of a repository
+      // List all branch names of a repository
       // -------
 
       this.listBranches = function(cb) {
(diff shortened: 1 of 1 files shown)


github.js
```diff
@@ -306,13 +306,33 @@
         });
       };
 
-      // List all branches of a repository
+      // List all branch names of a repository
       // -------
 
       this.listBranches = function(cb) {
```
glb888: Y

glb888: App

aendrew:
I'm not entirely sure what's going on with this PR; would somebody mind explaining please? Thanks!

PR #120, opened 2014-05-22
no CommonJS dependencies if run in browser

kpe:
I'm wandering if it wouldn't be better to reduce CommonJS dependencies, i.e. require('xmlhttprequest') and require('underscore') when running in a browser.

I'm using this awesome library from within TiddlyWiki, which runs in a browser and also uses kind of CommonJS requires.

P.S. sorry for my first pull request - it was build against a different fork, which used fewer underscore methods.

### DIFF
--- a/github.js
+++ b/github.js
@@ -17,9 +17,25 @@
   }else{
       // use native ECMAScript 5 if underscore is not available
       _ = window._ || (Array.prototype.map && Array.prototype.forEach && Array.prototype.slice && {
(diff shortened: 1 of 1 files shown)


aendrew:
I'm not sure of the benefit of this and I worry it might break things. Could you please rebase this? I'll then run the current test suite against it. Thanks!

aendrew:
This is irrelevant due to Underscore being removed. Closing as duplicate.

PR #121, opened 2014-06-24
Adds optional "force" parameter to method "write"

amenadiel:
Adds optional parameter ```force``` to the ```write``` method, in compliance with Github spec

https://developer.github.com/v3/git/refs/#update-a-reference

This allows to write a non fast forward content without getting error 422, (unprocessable entity)

### DIFF
--- a/github.js
+++ b/github.js
@@ -428,8 +428,8 @@
       // Update the reference of your head to point to the new commit SHA
       // -------
 
-      this.updateHead = function(head, commit, cb) {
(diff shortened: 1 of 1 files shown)


aendrew:
@amenadiel This looks great, could you please squash your commits? Thanks!

amenadiel:
I squashed, but this patch isn't mergeable to master anymore. I see that method "write" does no longer call "updateHead", so this PR is not needed anymore.

PR #127, opened 2014-10-09
Minor path change for the contents function

smakinson:
Changed the repo contents function to include the given path as part of the url so that it will return directories instead of always the root.

### DIFF
--- a/github.js
+++ b/github.js
@@ -445,7 +445,7 @@
       // --------
 
       this.contents = function(branch, path, cb, sync) {
-        return _request("GET", repoPath + "/contents?ref=" + branch + (path ? "&path=" + path : ""), null, cb, 'raw', sync);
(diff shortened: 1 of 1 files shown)


aendrew:
1. This duplicates the fix in #129 on ln. 418.
2. You then don't existence check "path" on ln. 391, and also the whitespace is messed up.

If you still think this is necessary, please write a ternary operation similar to the one done on `path` later in that line so an Uncaught ReferenceError exception isn't throw if the `path` argument isn't supplied to the function. Please also fix the whitespace issue, and remove the bit about `author` if #129 is merged by then. Lastly, squash into one commit. Thanks!

aendrew:
Also a duplicate of #92.

aendrew:
N.b., I've merged #92. If you wish this PR to be merged, please strip it back to just fixing the `author` issue — otherwise please close. Thanks!

aendrew:
Closing as duplicate.

PR #153, opened 2015-01-09
Adds bower.json

aendrew:
Would be nice to be able to manage this via Bower. You'll want to add it to the directory via [`bower register`](http://bower.io/docs/creating-packages/#register).

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,0 +1,25 @@
+{
+  "name": "github-api",
+  "main": "github.js",
+  "version": "0.10.3",
+  "homepage": "https://github.com/michael/github",
+  "authors": [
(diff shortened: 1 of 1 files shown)


eugef:
+1 for Bower support

mateusfreira:
+1 for Bower support

andrewmcwatters:
:+1:

onmyway133:
It is OK but I can't find it in http://bower.io/search/?q=github

clayreimann:
Bower support was deprecated because dist files don't belong in git. You can install the npm package for the built files, or you can use npmcdn if you don't want to host the files yourself.

PR #160, opened 2015-02-06
Fixes github.read

tmcw:
This was sending a JSON body with a GET request when it should
have been including the `ref` as a query parameter. This pull request
changes the behavior just for the `.read()` operation.


ray1007:
Had the same problem.
Thanks for this !

PR #185, opened 2015-05-24
Support Chinese file name

ele828:
As a Chinese, I can't upload a file with Chinese File name. Or maybe some other languages besides English.

Here, I encode the URI to be uploaded to make it works well.

thanks :)

### DIFF
--- a/github.js
+++ b/github.js
@@ -662,7 +662,7 @@
       this.write = function(branch, path, content, message, cb) {
         that.getSha(branch, path, function(err, sha) {
           if (err && err.error !== 404) return cb(err);
-          _request("PUT", repoPath + "/contents/" + path, {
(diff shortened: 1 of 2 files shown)


aendrew:
Hi! Thanks for your contribution!

Seems all the tests are passing, and the issue is more due to how I'm adding btoa in NodeJS. 

That said, could you please contribute some unit tests and rebase? Thanks!

ele828:
Of course, but unfortunately, I can't run the source code. After I run ``npm install``, all the test cases in test folder goes wrong, which is caused by ``btoa is not defined``. :(...

aendrew:
@ele828 That's been fixed in the last commit I've just pushed. Pull from `master` please!

ele828:
@aendrew It's done, works well :)

aendrew:
Cool! Thanks for your contribution!

Issue #45, opened 2013-03-27
Unit test needed to prevent #14 regression


runemadsen:
Hi there

Awesome library. I have one question though:

It seems like you're appending random query strings to the end of every API call, probably in order to make sure you get a non-cached response on every call. 

However, the Gihub API implements a pretty smart HTTP caching, that is being ignored with this library. For people using it, it means that the rate-limit will get used much faster, and apps will try to reload content that is not updated.

Here's an example of what I mean:

```javascript

/* Using JQuery
____________________________________________ */

$.ajax({
  url:"https://user:pass@api.github.com/repos/user/repo",
  success: function(data) { console.log(data) }
});
// --> response 200 with content

$.ajax({
  url:"https://user:pass@api.github.com/repos/user/repo",
  success: function(data) { console.log(data) }
});
// --> if called again within 60 seconds this call will never be made, 
//      as the Github API has HTTP cache of 60 seconds. If after 
//      60 seconds and data is not updated, it returns an empty 
//      304 not modified, which is much faster

/* Using github.js
____________________________________________ */

var github = new Github({
  username: "user",
  password: "pass",
  auth: "basic"
});
var repo = github.getRepo(user, repo);

repo.show(function(err, repo) { console.log(repo) });
// --> response 200 with content

repo.show(function(err, repo) { console.log(repo) });
// --> response 200 with content, no caching

repo.show(function(err, repo) { console.log(repo) });
// --> response 200 with content, no caching

```

I saw that you have "implemented smart caching" in a commit, so I'm not sure exactly what's going on. How is this library caching stuff?

aendrew:
Seriously? Lulz @ "smart caching"...

This should be fixed and the whole random number thing removed. Marking as P2 bug for 0.10.7.

aendrew:
I've removed the cache buster from _request. This means that it should fall back to XMLHttpRequest's caching layer. 

*Edit:* Apparently XMLHttpRequest only caches responses in the browser. This probably needs to be implemented in the lib. Reopening...

aendrew:
Actually, this is now resolved — GitHub only sends a 304 and no body if you send a conditional header. Therefore there's no reason to add a cache buster to the URL (other than to get around the browser's caching; if anything, the cache buster should be added via an option, though given I am currently unsure how to do that, am leaving it be for the moment).

[...]

PR #186, opened 2015-05-25
expose xhr

sielay:
Small change to expose XHR, so we can use conditional requests and save some quota ;)

### DIFF
--- a/github.js
+++ b/github.js
@@ -132,8 +132,8 @@
       // -------
 
       this.orgs = function(cb) {
-        _request("GET", '/user/orgs', null, function(err, res) {
-          cb(err, res);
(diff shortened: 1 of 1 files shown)


aendrew:
@sielay Excellent, would you mind writing a quite test suite for this? Would love to add this to 0.10.7 if possible! Apologies for the delay getting back to you!

aendrew:
I fixed the conflicts and merged this in 2ac1fb9153ccd11d5083f5669fb3653440581b83. I don't think it needs unit tests, it's just adding a callback argument. Thanks a bajillion @sielay — you've made working on #45 much easier! :smile:

sielay:
@aendrew thanks, forgot about this one already ;)

PR #197, opened 2015-06-04
Convert data in queryParams when GET or HEAD requests

### LINKED ISSUE #195: Bad use of the send() method of XMLHttpRequest
According to the [XMLHttpRequest whatwg.org specification](https://xhr.spec.whatwg.org/#the-send()-method):
`If the request method is GET or HEAD, set body to null.`

When use _request method, data was not send to the GitHup API:
https://github.com/michael/github/blob/master/github.js#L525
https://github.com/michael/github/blob/master/github.js#L599

It works well on master branch 'cause GitHub API set ref to master by default.

firehist:
It resolves #195 :rocket:

### DIFF
--- a/github.js
+++ b/github.js
@@ -48,7 +48,14 @@
     function _request(method, path, data, cb, raw, sync) {
       function getURL() {
         var url = path.indexOf('//') >= 0 ? path : API_URL + path;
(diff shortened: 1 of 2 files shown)


aendrew:
LGTM! Thanks for the tests!

firehist:
:fireworks: :)

PR #201, opened 2015-06-12
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3

Also their maintainer says they probably won't ever use it: http://stackoverflow.com/questions/24844901/bowers-bower-json-file-version-property

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,7 +1,6 @@
 {
   "name": "github-api",
   "main": "github.js",
-  "version": "0.10.6",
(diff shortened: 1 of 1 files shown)


aendrew:
Oh wicked awesome — I hate having to bump that with every release, always forget. Thanks! :+1: :shipit:

aendrew:
Ah, crap — I wish there was a "merge and comment" button, totally closed it. Will commit your changes and add credit. Thanks for this!

PR #202, opened 2015-06-22
Fixed commit on and read from a non-master branch

ctalau:
There was a problem with repo.write and repo.read from a branch that was not master. The ref GET parameter was passed in the message body. I fixed it and added a test-case.


aendrew:
LGTM! Thanks!

PR #203, opened 2015-06-22
Unicode content support

ctalau:
There were problems with the base64 encoding of Unicode characters. Fixed it to work both in browsers and in node.js.


aendrew:
Many thanks! Merged!

PR #204, opened 2015-06-30
Improve docs

pidofme:
Fix minor documentation mistake


aendrew:
Yup! Wow, wonder how long that was there for... :+1: Thanks!

Issue #107, opened 2014-03-05
remove underscore?


Pomax:
Underscore is used all of seven times in the codebase, mostly for functions that plain JS already has. It would be incredibly nice if these seven bits were actually turned into plain JS so that github.js would just work as a standalone library

aendrew:
Fixed via #207. Thanks, @ctalau!

PR #207, opened 2015-07-18
Removed the underscore.js dependency

ctalau:
This pull request is a fix for #107. It gets rid of the underscore.js dependency while preserving compatibility with IE9+.


aendrew:
Hero! Amazing! Thank you!

PR #212, opened 2015-08-19
Fix url query string mistakes `?&`

ilyaigpetrov:
1. No more `?&` in urls.
2. Now stamp precedes other params for programming convenience, but it doesn't make urls invalid.
3. `map`, `join`? ugh.. wouldn't it be easier to read `for`-loop?
4. `hasOwnProperty` added because `jshint` doesn't detect iteration through `Object.keys` and complains about it. You could fix `.jshint` as well, add `var param in Object.keys(data)` and get rid of `hasOwnProperty`.

### DIFF
--- a/github.js
+++ b/github.js
@@ -46,14 +46,14 @@
     function _request(method, path, data, cb, raw, sync) {
       function getURL() {
         var url = path.indexOf('//') >= 0 ? path : API_URL + path;
-        url += ((/\?/).test(url) ? '&' : '?');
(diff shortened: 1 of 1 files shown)


aendrew:
Merged in 61f526ff8c6124fae8635fd9b9b3e01810e2a7c8. Quite agreed that a map + join is suboptimal for simply building a querystring. Thanks, @ilyaigpetrov!

PR #215, opened 2015-08-29
Move filters out tree nodes, rather than delete their sha

mushishi78:
I was getting an error when moving files, and I came across [this issue](https://github.com/philschatz/octokit.js/issues/78) with a suggested workaround that worked for me. So I thought I'd send over a PR just incase.

However, moving files was working for me just fine before, and so the real cause of the problem is a mystery as far as I concerned. If no-one else is having the problem, then I understand if you don't feel a need to merge.

### DIFF
--- a/github.js
+++ b/github.js
@@ -655,9 +655,9 @@
         updateTree(branch, function(err, latestCommit) {
           that.getTree(latestCommit+"?recursive=true", function(err, tree) {
             // Update Tree
-            tree.forEach(function(ref) {
+            tree = tree.filter(function(ref) {
               if (ref.path === path) ref.path = newPath;
-              if (ref.type === "tree") delete ref.sha;
+              return ref.type !== "tree";
             });
 
             that.postTree(tree, function(err, rootTree) {


aendrew:
Hi @mushishi78! 

Could you please write a unit test for this? That would really help in determining whether this is a valid issue or not. I suggest writing the test case and running it against `master` HEAD and your PR; should fail in the former and pass in the latter. Thank you!

mushishi78:
Sorry to disappoint you again, but I also don't know how to write a test for this either. As far as I can tell, there is no previously written test for the move function, and even if their was one, I'd need to understand what caused the bug in order to write a test that isolated the bug.

To write a test for the move function, I'd think you'd need a to do some set up code for your testing that would create a test branch on a test repo and then delete the branch afterwards.  That way you'd be free to make changes to without repercussion, you could rely on the same files being present in the repo  and I could move some files or fail to move some files around with reckless abandon.

aendrew:
Bumping to 0.10.8, we really need to release 0.10.7.

aendrew:
Closed in error; reopening.

AurelioDeRosa:
@clayreimann I think this PR can improve the lib as it sends only the data needed. Should I rebase it?

PR #217, opened 2015-09-06
Add options argument for notifications()

yuya-takeyama:
For the backward compatibility, options can be omitted


aendrew:
:100: Full marks! Many thanks!

PR #220, opened 2015-09-17
Clarify docs for 'username' parameter in user.show

arosenberg01:
- The docs are unclear about how to retrieve user information for the currently authenticated user via GET '/user' vs GET /users/:username
- Useful for identifying users via Github OAuth


aendrew:
Cool, thanks!

PR #194, opened 2015-06-01
Modified Github constructor, getCommits, and getCommit. Added getComments, createComment, and updateComment.

lukeburns:
### Modifications:
- options parameter for the Github constructor is now optional, so you can instantiate with `new Github()` to access public data without specifying authorization information.
- `getCommits` was failing when only passed a callback. options parameter is now optional.
- corrected path (should not contain '/git/') and removed unnecessary argument `branch` from `getCommit(sha, cb)`.
- I removed the `branch` parameter from the `getCommit` test, because the Github API does not take a `branch` parameter.

### Added:
- Get commit comments: `getComments([sha,] cb)`
- Create a commit comment: `createComment(sha, body, [options,] cb)`
- Update a commit comment: `updateComment(id, body, cb)`
- Added tests for `getCommits` and `getComments`

PR #221, opened 2015-09-20
Expose request methods.

adjohnson916:
Simple change to expose request methods. I'd like to be able to use them for unsupported API requests like to get issue comments (#194). 

I've left them underscore-prefixed, as they are ancillary and their existence or behavior may change in the future.

Especially useful is the `_requestAllPages` with `Link: rel="next"` handling. If anyone is aware of an external module that does that, then I won't need this. But otherwise, perhaps these could be extracted into an external module, or replaced by some other existing external module.

### DIFF
--- a/github.js
+++ b/github.js
@@ -43,7 +43,7 @@
     //
     // I'm not proud of this and neither should you be if you were responsible for the XMLHttpRequest spec.
 
-    function _request(method, path, data, cb, raw, sync) {
+    var _request = Github._request = function _request(method, path, data, cb, raw, sync) {
       function getURL() {
         var url = path.indexOf('//') >= 0 ? path : API_URL + path;
         url += ((/\?/).test(url) ? '&' : '?');
@@ -92,9 +92,9 @@
       if (sync) {
         return xhr.response;
       }
-    }
+    };
 
-    function _requestAllPages(path, cb) {
+    var _requestAllPages = Github._requestAllPages = function _requestAllPages(path, cb) {
       var results = [];
       (function iterate() {
         _request('GET', path, null, function(err, res, xhr) {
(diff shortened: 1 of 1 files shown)


raszi:
:+1:

aendrew:
@adjohnson916 Awesome, thanks! Mind adding a unit test or two? Happy to merge afterwards. :smile:

adjohnson916:
@aendrew Sure, what kind of tests would you want? I believe existing test cases implicitly depend on this function, so to add more might mean double coverage. Would you just want a test that merely guarantees the function is externally available? If it's to become part of the public API contract, would you prefer to remove the underscore prefix?

AurelioDeRosa:
I agree on removing the underscore as it'll become public. I also think that at this stage we don't want to update all the other methods that call `_request`, so for now it's better to stick with the current signature.

In regard of the tests, I think that checking that the feature is available externally is enough. As you said, this function is already covered by all the other tests.

The only change I'd make is that I wouldn't introduce an additional variable but rather split the statement. So, leave the function as it is and have at the bottom of the file something like:

```
Github.request = request;
```

AurelioDeRosa:
@aendrew what do you think about my suggestion? If you agree, we can move forward with this idea.

aendrew:
@AurelioDeRosa Apologies, didn't see the activity on this PR until now. That sounds good, go for it! :+1:

AurelioDeRosa:
I had a second thought about my change. Splitting the statement would move the second one very down the code. Merged as is.

aendrew:
Cool! :+1:

adjohnson916:
@AurelioDeRosa Thanks. I was waiting on word about the unit tests. Also, perhaps the `var` assignments are superfluous. Oh well!

raszi:
@AurelioDeRosa when are you planning to release this?

PR #223, opened 2015-09-24
repo: add getRepo with fullname

yorkie:
Hooks API has direct full_name to be respond, so adding a way to
initistant the repo object with full_name should be more convinent.

Now we are able to get a repo:

```js
githubapi.getRepo('nodejs/node')
```

Thanks

### DIFF
--- a/github.js
+++ b/github.js
@@ -235,9 +235,16 @@
     Github.Repository = function(options) {
       var repo = options.name;
       var user = options.user;
+      var fullname = options.fullname;
 
       var that = this;
-      var repoPath = '/repos/' + user + '/' + repo;
+      var repoPath;
+
+      if (fullname) {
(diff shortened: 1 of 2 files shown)


EliCDavis:
:+1: , would have helped a lot on my project earlier :sweat_smile:

yorkie:
Hey @EliCDavis if you hope directly apply my changes into your project, you are able to use https://github.com/weflex/octop for now, because I found the repo under DevelopmentSeed still not active and with so many PR and issues without any feedback. Because I will be working on Github API, so I think Octop might be more active during this time, BTW, of course hope this would be merged as well :-)

aendrew:
This looks like a great commit. Thanks for adding unit tests! Merging for 0.10.7!

yorkie:
:-)

PR #225, opened 2015-09-29
Fix to allow both writing new files and overwriting old

JensLincke:


aendrew:
Brilliant, could you please add a unit test and rebase? Thanks!

aendrew:
That's actually a really nice fix; I don't think it needs a unit test because it would break other tests were it not. Merging now.

Issue #31, opened 2012-12-03
listBranches() only returns the last component of the branch path


timrchavez:
After looking at the code, it looks like this was done intentionally, but I'm not sure why.  I feel I should be able to pass the values returned by listBranches() to getRef() (appended to "heads/") without issue, but if I have a branch 'feature/foo', listBranches() will return 'foo', and getRef() will fail.

olivierbeaulieu:
Fixed [here](https://github.com/michael/github/pull/231)

aendrew:
Fixed by #231. Thanks, @olivierbeaulieu!

PR #231, opened 2015-10-07
Allow repo.listBranches to support branch names that contains slashes

olivierbeaulieu:
I was running into an issue with `listBranches` - we use slashes in our branch naming convention.

For example, the branch `feature/my-cool-feature` would return only `my-cool-feature`. This fixes this issue.

### DIFF
--- a/github.js
+++ b/github.js
@@ -12,7 +12,7 @@
 
 (function() {
   'use strict';
-  
+
   // Initial Setup
   // -------------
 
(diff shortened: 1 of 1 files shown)


aendrew:
See #31.

aendrew:
Hmm, tests failed in Node 0.11 but not 0.10. I'd love to add this to 0.10.7, will take a look at fixing your tests.

aendrew:
Phh, what the heck? Now tests pass fine. MERGING. :+1: :shipit: 

Many thanks, @olivierbeaulieu!

PR #240, opened 2015-10-21
Add type, sort, per_page, and page options to User.repos call

mikec:
I needed the ability to set these

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,10 +215,10 @@ repo.contributors(function(err, data) {});
 var user = github.getUser();
 ```
 
-List all repositories of the authenticated user, including private repositories and repositories in which the user is a collaborator and not an owner.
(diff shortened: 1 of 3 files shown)


aendrew:
Hi @mikec!

Could you please add a unit test for these options? Will happily merge after. Thanks!

mikec:
@aendrew paging wasn't working so I updated `user.repos` to get a single page instead of all pages. I think the same should be applied to `user.userRepos` and `user.orgRepos`

aendrew:
@mikec Quite agreed re: `user.userRepos` and `user.orgRepos`, nice implementation. I'll merge this now, could you do userRepos and orgRepos as a separate PR? Thanks!

mikec:
@aendrew :+1:

aendrew:
Merged! Thanks for your excellent contribution! :+1: :shipit:  :tada:

Issue #243, opened 2015-10-23
Decide upon community code standards


aendrew:
As we move towards rewriting the lib in modern ECMAScript, we should probably decide upon some standards beyond what's specified with jsHint. JSHint itself is moving towards being more of a code sanity checker and less of a code formatter, which is why JSCS has popped up. I'm also a fan of clang-format, which is nearly zero-config and works brilliantly.

To that end, we should probably decide the following things:
* JSCS or clang-format?
* Spaces vs. tabs
* 4 vs. 2 indentation
* double or single quotes?
* other stuff?

My personal preference is for clang-format + tabs + 2 indentation + single quotes. Please comment if you prefer something otherwise. N.b., I reserve the right to lock this thread at any time if it gets too bikeshed-y.

AurelioDeRosa:
Here are my thoughts:

- I'd go with JSCS as it has an awesome community. It also has [a tool for jsDoc](https://github.com/jscs-dev/jscs-jsdoc) and integration with Grunt
- Spaces
- [3 spaces](http://www.audero.it/blog/2015/10/21/the-revolution-of-3-spaces-code-indentation/)
- Single quotes

aendrew:
What if we used JSCS for jsDoc linting and clang-format for auto-formatting projects? I'm almost even on board with the 3-spaces idea, though I worry that less experienced contributors will screw it up. *However*, if we use something that's able to reliably format code to standards, anyone can use anything they want and clangformat will correct it to whatever we decide upon as a standard.

AurelioDeRosa:
In regard of your point about how many spaces, I think we'll definitely fix everything using a tool. I don't mind using something other than JSCS. Does is it have the some amount of options? I'm asking because I don't have experience with it. Just a FYI, JSCS has a fix option as well.

AurelioDeRosa:
What we should use for tooling? Grunt, Gulp, or simply npm?

aendrew:
I'm aware JSCS has a fix option, but have no idea how accurate or effective it is. Granted, I don't have a ton of experience with clangformat either; autoformatting the code on commit is slightly new territory for me. See [this](http://clangformat.com/) for .clangformat options.

re: tooling — for the 0.10.x series, if we're going to use Karma, it's probably worthwhile setting up some sort of task runner. Happy to set up some Gulp tasks, wouldn't take long at all. Gasp, we could even have a minified version that way! :sweat_smile: 

[...]

PR #244, opened 2015-10-23
Adds .clang-format file. See #243.

aendrew:
As proposed in #243, here's a really basic `.clang-format` if we do decide to go the clang route.


aendrew:
Forget it, I want something that can fix inconsistent quotation marks.

PR #248, opened 2015-10-26
Rewrote all unit tests to use Mocha Chai instead of Tape. Fixes #237.

### LINKED ISSUE #237: Replace Tape with Karma + (Mocha, Jasmine or Qunit)
Testling has apparently been borked for a year. It should be removed, especially because we're using Travis for CI anyway. See substack/testling#88.

aendrew:
This removes Tape and replaces it with Mocha + Chai + Sinon. Sinon might be unnecessary because we're testing against a real username/repo most of the time.

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -1,5 +1,5 @@
 {
-   "disallowDanglingUnderscores": true,
+   "disallowDanglingUnderscores": false,
    "disallowIdentifierNames": [],
    "disallowImplicitTypeConversion": [],
    "disallowKeywordsOnNewLine": [
@@ -53,7 +53,6 @@
       "do",
       "else",
       "for",
-      "if",
       "try",
       "while"
    ],
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -1,0 +1,49 @@
+module.exports = function(config) {
+   'use strict';
+
+   var configuration = {
+      client: {
+         captureConsole: true
+      },
+
+      singleRun: true,
+
+      autoWatch: false,
+
+      frameworks: ['mocha', 'chai'],
+
+      browsers: ['PhantomJS'],
+
+      coverageReporter: {
+         reporters: [
+            {
+              type: 'lcov'
+           },
+           {
+              type: 'text-summary'
+           }
+         ],
+         instrumenterOptions: {
+            istanbul: {
+               noCompact: true
+            }
+         }
+      }
+   };
+
+   // This block is needed to execute Chrome on Travis
+   // If you ever plan to use Chrome and Travis, you can keep it
+   // If not, you can safely remove it
+   // https://github.com/karma-runner/karma/issues/1144#issuecomment-53633076
+   if (configuration.browsers[0] === 'Chrome' && process.env.TRAVIS) {
+      configuration.customLaunchers = {
+         'chrome-travis-ci': {
+            base: 'Chrome',
+            flags: ['--no-sandbox']
+         }
+      };
+      configuration.browsers = ['chrome-travis-ci'];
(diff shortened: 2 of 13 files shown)


AurelioDeRosa:
I suggest that  we add the gulpfile to the linter.

package.json
```diff
@@ -17,13 +19,23 @@
...
+    "karma": "^0.13.14",
+    "karma-chai": "^0.1.0",
+    "karma-coverage": "^0.5.3",
+    "karma-mocha": "^0.2.0",
+    "karma-mocha-reporter": "^1.1.1",
+    "karma-phantomjs-launcher": "^0.2.1",
+    "karma-sinon": "^1.0.4",
+    "mocha": "^2.3.3",
+    "phantomjs": "^1.9.18",
     "plato": "^1.4.0",
```
AurelioDeRosa: Are we still using this Plato?

aendrew: @AurelioDeRosa I thought I removed that. Not at the moment, but it might be good to add later. Agreed with adding Gulpfile to linter, will update commit with both fixes.

aendrew:
@AurelioDeRosa Quite agreed; have added all JS files and linted them all.

.jscsrc
```diff
@@ -38,7 +38,7 @@
    "disallowYodaConditions": true,
    "maximumLineLength": 120,
    "requireBlocksOnNewline": true,
-   "requireCamelCaseOrUpperCaseIdentifiers": true,
+   "requireCamelCaseOrUpperCaseIdentifiers": false,
```
AurelioDeRosa: What are the exceptions in our naming conventions to disable this? Just wondering.

aendrew: Main reasons I disabled them were _request and  (dangling underscores) and "per_page" from GitHub's API (camelcase or upper).

AurelioDeRosa: I see, good point. I think that in regard of `_request` we can rename it wherever it's used because it isn't even private anymore. In regard of page, we can add an _ad hoc_ exception on a specific line as we do with Istanbul. What do you think?

aendrew: @AurelioDeRosa Yep, agreed on renaming _request and re-enabling the check for dangling underscores. That's also a good suggestion re: ad hoc overriding where-ever GitHub uses an underscore for a space, will help prevent people from naming things badly. 

For the moment though, can we leave _request as-is and save that for a separate PR? This one is already getting a bit unwieldy. :smile: Will update to re-enable requireCamelCase.

AurelioDeRosa: Sounds good.

karma.conf.js
```diff
@@ -0,0 +1,58 @@
...
+
+      singleRun: true,
+
+      autoWatch: false,
+
+      frameworks: ['mocha', 'chai'],
+
+      browsers: ['PhantomJS'],
+
+      plugins: [
```
AurelioDeRosa: This can be removed as mentioned in the documentation:

> By default, Karma loads all sibling NPM modules which have a name starting with karma-*.

test/test.issue.js
```diff
@@ -1,36 +1,46 @@
...
-
-
-});
\ No newline at end of file
+var testUser, github, issues;
+
+if (typeof window === 'undefined') {
+   // Module dependencies
+   var chai = require('chai');
+   var Github = require('../');
```
AurelioDeRosa: I think we should explicitly call the file. This will be updated when we'll split in modules.

aendrew:
Closed by 1813b0d0965d460158fbeec4f2ad3d7a9becd9ba.

PR #268, opened 2015-12-29
When getting all pages, if the API response is an object, wrap it in an array before pushing it to results.

### LINKED ISSUE #267: Support endpoints that return objects in _requestAllPages
Currently, if you call `_requestAllPages` on an endpoint that returns an object, like `/search/`, you won't get any results back.

This is due to the handling of the result object - it's assumed it will be an array.

See: https://github.com/michael/github/blob/master/github.js#L124

jeffstieler:
Fixes #267.

### DIFF
--- a/github.js
+++ b/github.js
@@ -121,8 +121,8 @@
                   return cb(err);
                }
 
-               if (Object.prototype.toString.call(res) === '[object Object]') {
-                  res = [ res ];
+               if (!(res instanceof Array)) {
+                  res = [res];
                }
 
                results.push.apply(results, res);
--- a/github.js
+++ b/github.js
@@ -121,6 +121,10 @@
                   return cb(err);
                }
 
+               if (Object.prototype.toString.call(res) === '[object Object]') {
+                  res = [ res ];
+               }
+
                results.push.apply(results, res);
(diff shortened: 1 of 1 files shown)


github.js
```diff
@@ -121,6 +121,10 @@
                   return cb(err);
                }
 
+               if (Object.prototype.toString.call(res) === '[object Object]') {
```
AurelioDeRosa: Is there any reason why you haven't used:

```js
if (typeof res === 'object')
```

jeffstieler: `typeof` returns `'object'` for arrays, see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof

AurelioDeRosa: Right, I didn't think about it. A better suggestion is:

```js
if (!(res instanceof Array))
```

AurelioDeRosa:
Hi @jeffstieler. Would you mind update your PR?

jeffstieler:
@AurelioDeRosa sorry this slipped through the cracks - updated.

AurelioDeRosa:
Merged.  Thank you.

Issue #278, opened 2016-01-13
pull request merge


benjamine:
is there a method to merge a pull request?
(if not, would you be interested on adding it?)

https://developer.github.com/v3/pulls/#merge-a-pull-request-merge-button

AurelioDeRosa:
Hi @benjamine.

We don't have such method but we're interested in adding it. Would you mind submit a PR?

benjamine:
cool, yes, I'll give it a try

AurelioDeRosa:
FYI,  in case you want this plus some other method , you can take a look at [this project of mine called GitHub Extended](https://github.com/github-tools/github-extended#methods)

benjamine:
oh interesting, already sent the PR to add it here, but I'll take a look.

PR #281, opened 2016-01-20
merge pull request

benjamine:
adding `repo.mergePull`
docs: https://developer.github.com/v3/pulls/#merge-a-pull-request-merge-button

note the unit test isn't really testing a succesful merge (but a failture to merge), given the destructive nature of that action I think that would require mocking of the github API, but that might be a big effort.

ps: this is a pull request to merge pull requests :open_mouth:

### DIFF
--- a/test/test.repo.js
+++ b/test/test.repo.js
@@ -33,6 +33,7 @@ if (typeof window === 'undefined') { // We're in NodeJS
       // jscs:disable
       imageB64 = 'iVBORw0KGgoAAAANSUhEUgAAACsAAAAmCAAAAAB4qD3CAAABgElEQVQ4y9XUsUocURQGYN/pAyMWBhGtrEIMiFiooGuVIoYsSBAsRSQvYGFWC4uFhUBYsilXLERQsDA20YAguIbo5PQp3F3inVFTheSvZoavGO79z+mJP0/Pv2nPtlfLpfLq9tljNquO62S8mj1kmy/8nrHm/Xaz1930bt5n1+SzVmyrilItsod9ON0td1V59xR9hwV2HsMRsbfROLo4amzsRcQw5vO2CZPJEU5CM2cXYTCxg7CY2mwIVhK7AkNZYg9g4CqxVwNwkNg6zOTKMQP1xFZgKWeXoJLYdSjl7BysJ7YBIzk7Ap8TewLOE3oOTtIz6y/64bfQn55ZTIAPd2gNTOTurcbzp7z50v1y/Pq2Q7Wczca8vFjG6LvbMo92hiPL96xO+eYVPkVExMdONetFXZ+l+eP9cuV7RER8a9PZwrloTXv2tfv285ZOt4rnrTXlydxCu9sZmGrdN8eXC3ATERHXsHD5wC7ZL3HdsaX9R3bUzlb7YWvn/9ipf93+An8cHsx3W3WHAAAAAElFTkSuQmCC';
       imageBlob = new Blob();
+
       // jscs:enable
    }
 }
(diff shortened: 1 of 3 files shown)


benjamine:
from #278

github.js
```diff
@@ -458,6 +458,26 @@
             });
          };
 
+         // Merge a specific pull request
+         // -------
+
+         this.mergePull = function(pull, message, cb) {
```
AurelioDeRosa: Since we're using already an object as the first parameter, instead of using  overloading I think it's better to add the optional message as a property of the object. If it isn't defined we can use a default message (e.g. empty string).

github.js
```diff
@@ -458,6 +458,26 @@
...
+               cb = message;
+               message = '';
+            }
+
+            var data = {
+               sha: pull.sha || pull.head.sha,
+               commit_message: message
+            };
+
+            _request('PUT', repoPath + '/pulls/' + pull.number + '/merge', data, function(err, mergeResult, xhr) {
```
AurelioDeRosa: The anonymous function can be replaced by `cb`.

test/test.repo.js
```diff
@@ -272,14 +273,35 @@ describe('Creating new Github.Repository', function() {
    it('should get pull requests on repo', function(done) {
       var repo = github.getRepo('michael', 'github');
 
-      repo.getPull(153, function(err) {
+      repo.getPull(153, function(err, pull) {
```
AurelioDeRosa: Can you remove this change and move it into another PR? This is not related to the merge feature.

AurelioDeRosa:
Hi @benjamine. Are you still interested in this PR?

benjamine:
@AurelioDeRosa yes, fixing

benjamine:
@AurelioDeRosa all fixed except for this comment that got lost while rebasing:
> Since we're using already an object as the first parameter, instead of using overloading I think it's better to add the optional message as a property of the object. If it isn't defined we can use a default message (e.g. empty string).
the reason for I added `message` as a separate (optional) argument is that this allows you to send as first paramenter a `pull` object obtained from `getPull` (which will probably the most common case, as you can't create that yourself, github forces you to send the head sha to avoid mistakes)
This is shown in the 2nd example I added to README.
Otherwise you'd have to modify the object in the middle (with a parameter that isn't really part of a pull request info).

test/test.repo.js
```diff
@@ -425,6 +426,25 @@ describe('Creating new Github.Repository', function() {
       });
    });
 
+   it('should merge pull requests on repo', function(done) {
+      var repo = github.getRepo('michael', 'github');
+      var expectedDocUrl = 'https://developer.github.com/v3/pulls/#merge-a-pull-request-merge-button';
+
+      repo.getPull(153, function(err, pull) {
+         repo.mergePull(pull, function(err) {
```
AurelioDeRosa: I think we should have a test for the success of the merge of a pull request, not the failure. At the moment the test doesn't verify if the method is working correctly.

AurelioDeRosa:
I see your point @benjamine and it makes perfect sense. However, I know that in that context a user will not care about the change of the original object with an optional `message` property. Also, this will help me in the big method refactor I'm planning.

test/test.repo.js
benjamine: I agree, but given there's no mocking of the github API, I don't see a way to do merges in a repeatable, parallel-safe, way.
But maybe the test title could be negative then.

benjamine:
besides side effects, it seems confusing to me, as user, to add a `pull.message = ...` when it's not PR message you're setting (it's a commit message for the merge of such PR).
unless it's called `pull.mergeCommitMessage`?

test/test.repo.js
AurelioDeRosa: At the moment we're not doing parallel tests anyway, so the idea here is to create a new PR and then merge it.

benjamine: I meant parallel in that any user could be running them, locally, or in travis from their forks.
should this create a random commit to a new branch and merge it (with the potential to get conflicts if other is merged in between)?

AurelioDeRosa: No conflict can happen as the repository used is different at every execution of the tests.

benjamine: oh ok, good to know, well I'll copy from the create pull request test

Issue #229, opened 2015-10-06
Include several files in one single commit


danielo515:
Hello,

I can see there is no support for commiting several files on one go. Or at least it is not documented. Is this possible ?

Regards

mattpass:
I've looked into this and don't believe the GitHub API makes it possible but would love this also.

aendrew:
This is probably a lower-priority item but PRs welcome.

AurelioDeRosa:
Having worked with the API, I think this isn't possible at the moment. Can someone else confirm? If confirmed this issue can be closed for now and maybe opened again when the API will allow for such feature.

mattpass:
Went thru all their API docs about 9+ months ago and didn't seem to be any way of doing this. Also contacted GH dev team who didn't seem to want to confirm their API couldn't support multi-file commits. 

If this issue is to be closed due to non-support in the API, I think we should all ask GH team to make it possible, as it's a fairly significant and common thing surely?

mattpass:
Have just contacted GH via https://github.com/contact to again ask for multi-file commits. Maybe if other people do they'll take notice. :-)

mattpass:
Wow, OK, they got back to me within 3 minutes with the answer:

```
That's definitely possible, Matt -- you just need to use the low-level Git Data API:

https://developer.github.com/v3/git/

Here's what you do:

1. Create blobs for the three files
2. Create a tree which defines the folder structure for the three files
3. Create a commit for the tree
4. Update your branch to point to the new commit

Hope this helps and let us know if you run into problems.
```

Guess we know what needs to be done now to make this happen.

timbur:
Good info.  Thanks!

danielo515:
Hello @mattpass 

Thank you very much for the info. Since this library is an "high level wrapper for github API" is there any plan to wrap this feature?

regards

AurelioDeRosa:
If the other maintainers agree, I'll probably give it a go in the next weeks.

mattpass:
.@danielo515 - sounds like @AurelioDeRosa will give it a go. I think it's a pretty crucial aspect to cover and now we know what's involved, the above processes could certainly be wrapped up into a function.

aendrew:
@AurelioDeRosa Please go for it (and thanks for picking it up)! Am marking for 0.10.8, but am happy to add it to 0.10.7 if done in time.

gmaclennan:
[...]

PR #283, opened 2016-01-27
Add base_tree param in postTree. Fixes #282

### LINKED ISSUE #282: .postTree does not support base_tree
According to the Github API doc (https://developer.github.com/v3/git/trees/#create-a-tree) creating a new tree should support providing a `base_tree` param, so that you don't have to manually update your tree.

This is extremely important to have.

vmasto:
This change is absolutely necessary for any serious tree manipulation. 

It's also a requirement for a second PR I'm preparing which allows for multiple files committed in a single batch, as described in #229. I have the code for this ready but it needs polishing and testing before I can submit the PR.


clayreimann:
@vmasto Check out the [create tree](http://michael.github.io/github/docs/1.2.1/Repository.html#createTree) method, as I think it addresses this issue.

PR #284, opened 2016-01-28
Added low level access to edit and get an issue

marvinosswald:


README.md
```diff
@@ -382,6 +382,27 @@ To comment in a issue
 issues.comment(issue, comment,function(err, comment) {});
 ```
 
+To edit an issue
+
+| Parameter | Type |
```
AurelioDeRosa: We don't use this format in the README. Can you align it?

README.md
```diff
@@ -382,6 +382,27 @@ To comment in a issue
...
+| issue | object/int |
+| [options](https://developer.github.com/v3/issues/#parameters-3) | object |
+
+```js
+issues.edit(issue,options,function (err,issue) {});
+```
+
+To get an issue
+
+| Parameter | Type |
```
AurelioDeRosa: Same here

bower.json
```diff
@@ -23,5 +23,8 @@
     "bower_components",
     "test",
     "tests"
-  ]
+  ],
+  "dependencies": {
```
AurelioDeRosa: axios is not the only dependency we have. Take a look at our `package.json` file.

src/github.js
```diff
@@ -964,6 +964,18 @@
                body: comment
             }, cb);
          };
+         this.edit = function(issue,options,cb){
```
AurelioDeRosa: The snippet doesn't follow our code convention style. Can you use our `lint` task so that it will automatically fix it for you?

AurelioDeRosa:
Thank you for the PR. I've added a few comments for you.

AurelioDeRosa:
Thank you for the updates @marvinosswald. I'll review and merge it as soon as possible.

test/test.issue.js
```diff
@@ -32,4 +32,24 @@ describe('Github.Issue', function() {
          });
       });
    });
+   it('should edit the issues title', function(done){
```
AurelioDeRosa: The indentation of this code is incorrect. You have 2 spaces but we use 3.

test/test.issue.js
```diff
@@ -32,4 +32,24 @@ describe('Github.Issue', function() {
...
+         issues.edit(issuesList[0], {
+             title: 'Edited test title'
+         }, function(err, res) {
+            should.not.exist(err);
+            res.title.should.equal('Edited test title');
+            done();
+         });
+      });
+   });
+   it('should edit the issues title', function(done){
```
AurelioDeRosa: The title of this test is incorrect. It should say "should get issue".

test/test.issue.js
```diff
@@ -32,4 +32,24 @@ describe('Github.Issue', function() {
...
+       issues.list({}, function(err, issuesList) {
+         issues.get(issuesList[0], function(err, res) {
+            should.not.exist(err);
+            res.title.should.equal('Edited test title');
```
AurelioDeRosa: The test should assert that the number is correct, not be based on the previous test. Tests must always be independent.

src/github.js
```diff
@@ -964,6 +964,22 @@
                body: comment
             }, cb);
          };
+         this.edit = function (issue,options,cb) {
```
AurelioDeRosa: The indentation is incorrect. You used 4 spaces but we use 3.

src/github.js
```diff
@@ -964,6 +964,22 @@
...
+             if (typeof issue === 'object') {
+                 issue = issue.number;
+             };
+             _request('PATCH', path + '/' + issue, options, function(err, res) {
+               cb(err, res);
+            });
+         };
+         this.get = function (issue,cb) {
+             if (typeof issue === 'object') {
```
AurelioDeRosa: There is no need to accept an object here. The method retrieves an issue by its number, so the should accept a number as its first argument.

src/github.js
```diff
@@ -964,6 +964,22 @@
...
+                 issue = issue.number;
+             };
+             _request('GET', path + '/' + issue,{}, function(err, res) {
```
AurelioDeRosa: Here you can pass `cb` as the last argument. There is no need to define an anonymous function.

src/github.js
```diff
@@ -964,6 +964,22 @@
                body: comment
             }, cb);
          };
+         this.edit = function (issue,options,cb) {
+             if (typeof issue === 'object') {
+                 issue = issue.number;
+             };
+             _request('PATCH', path + '/' + issue, options, function(err, res) {
```
AurelioDeRosa: Here you can pass `cb` as the last argument. There is no need to define an anonymous function.

src/github.js
```diff
@@ -964,6 +964,22 @@
                body: comment
             }, cb);
          };
+         this.edit = function (issue,options,cb) {
+             if (typeof issue === 'object') {
```
AurelioDeRosa: In this case the first argument should be a number, it shouldn't be derived from the options.

bower.json
```diff
@@ -23,5 +23,11 @@
     "bower_components",
     "test",
     "tests"
-  ]
+  ],
```
AurelioDeRosa: While this change is correct, it doesn't belong to this PR as it addresses a different issue.

Issue #239, opened 2015-10-19
Rewrite inline comments to make use of jsDoc


aendrew:
jsDoc is terrific and well-supported, makes reading the code itself easier and allows the creation of full API docs with trivial effort. We should endeavour to rewrite all inline comments using jsDoc.

michael:
:+1: we also recently starting generating docs for Substance using jsDoc. 

E.g. http://substance.io/docs/module-ui_surface.ContainerEditor.html

We also made a little script that reacts to a web hook to trigger regeneration of docs on every commit to master.

However, for Substance there's a long way to go until we have each module covered.  Github.js as a single-file could be documented quite fast.

Thanks to @aendrew et al. for turning Github.js into a professional test-covered project! You really do a great job!

AurelioDeRosa:
100% agree on jsDoc.

clayreimann:
To Do:
* [ ] Evaluate themes
  * [ ] [Documentation Jekyll](http://idratherbewriting.com/documentation-theme-jekyll/)
  * [ ] [minami](https://github.com/nijikokun/minami)
* [ ] Auto publish to [gh-pages](https://www.npmjs.com/package/gh-pages)

clayreimann:
Closed by e0666fd, opening new for auto-publishing

Issue #285, opened 2016-02-06
Newbie question... how can I avoid "TypeError: Base64 is undefined"


dalareo:
I am learning to use the library, in order to get data from Github and build a dynamic website but...

what am I doing wrong?? This is my html sandbox file:

    <!doctype html>
    <html lang="en">
    <head>
    <meta charset="utf-8">
    <title></title>
    <meta name="description" content="">
    <meta name="author" content="">
    <link rel="stylesheet" href="">
    </head>
    <body>
    <script src="github.js" type="text/javascript"></script>
    <script src="http://code.jquery.com/jquery-latest.min.js" type="text/javascript"></script>
    <script type="text/javascript">
    var github = new Github({
    username: 'dalareo',
    password: '**********',
    reponame: 'wemooc',
    auth: 'basic'
    });
    var user = github.getUser();
    user.orgs(function(err, orgs) {});
    </script>
    </body>
   </html>

AurelioDeRosa:
Hi.

If you don't want to use a module loader such as Browserify or RequireJS, you need to use the file called `github.bundle.min.js` inside the `dist` folder. By the way, I'm closing this issue as it isn't a bug with the library.

dalareo:
Thanks @AurelioDeRosa

enki-code:
@AurelioDeRosa Including that information in the Read Me would be extremely helpful.. Having to search through closed issues to find out why my implementation wasn't working is troublesome.

I too didn't use a module loader, but wanted to test this out in jsFiddle, and kept getting axios is not defined.

Would you mind adding something about this to the Installation section?

PR #290, opened 2016-02-21
Use babel umd

clayreimann:
~~This should not be merged until Babel `6.6.0` is released.~~ `babel@6.6.x` is released.

This could constitute the beginnings of the rework for the `0.12.x`/`1.0.0` release

To Do:
- [x] Add test coverage back
- ~~Add browser tests with [zuul](https://www.npmjs.com/package/zuul)~~ *Will be fixed by* #300
- ~~Add deploy scripts that generate the `dist/` artifacts~~ *Will be fixed by* #299
- ~~Add auto-generated documentation~~ *Will be fixed by* #239


AurelioDeRosa:
@clayreimann the reason why we didn't use Babel anymore is because we'll lose most of the code coverage tooling. While it's definitely possible to have the final metric value (e.g. 85% covered), the covered/uncovered lines will be lost. Also, altought very painful, we still need the dist files as demonstrated by issues like https://github.com/michael/github/issues/285.

clayreimann:
I have:

* Moved the testing from `karma+mocha` to just `mocha`
* Fixed the tests to that they will actually run in `mocha` when I checkout from master all I see are timeouts when I try and run the tests
* Moved to using `babel`'s UMD wrapper to clean up our code
* Added a fix for how we handle Blobs in node vs the browser

If this all looks good I'll add one more commit that fixes indentation in `github.js`. I haven't done that yet to keep the diff simple and not muck everything up with whitespace changes.

I think the changes here are going to be critical in allowing us to modularize our codebase going forward.

aendrew:
This is a huge amount of work, @clayreimann — thank you very much for it. I've admittedly been a bit distant from how @AurelioDeRosa has been managing the code the last while so it's ultimately up to him, but having gone through most of the PR (albeit fairly quickly), it looks pretty terrific. 

1. It seems all the Sauce Labs stuff has been stripped out. Thank you and good riddance; how I had it set up was woefully broken and any benefit gained from it was outweighed by the additional hassle it caused. It's better to not have it and then set it up later after the codebase has stabilized a bit. So, :+1: from me on that.

2. `npm test` is broken (at least for the version of Gulp I have); either the Gulpfile should be updated so that `gulp test:mocha` is the default task of `test`, or `package.json` updated so that `npm test` starts the test cycle. I'm impartial to either solution.

3. There's still a karma.conf.js even though Karma isn't a dep (unless I'm missing something, it's probably okay to just remove that too). With Karma gone, is there any way (or point) to still test in the browser?

That's all for now, will add more if I see anything else.

clayreimann:
@aendrew Thanks for the review. I do plan on helping modularizing the library, but after #264 didn't go anywhere I figured it would be better to do it piecemeal. 

Other changes in a7587fd:
* Fixed `npm test`
* Fixed whitespace linting in `github.js`

.gitignore
```diff
@@ -1,6 +1,8 @@
 .DS_Store
 .idea
 node_modules/
-npm-debug.log
 coverage/
+
+.zuulrc
```
AurelioDeRosa: What's this file for?

gulpfile.js
```diff
@@ -1,185 +1,113 @@
...
+      .pipe(gulp.dest('dist'))
+      ;
 });
 
 gulp.task('default', ['clean'], function() {
-   gulp.start('lint', 'test', 'build');
+   gulp.start('lint', 'test:mocha', 'build');
 });
-
-var sauceLaunchers = {
```
AurelioDeRosa: Does the PR remove sauce lab completely? I remember we discussed  this point, but now I can't remember the reason on top of my mind.

package.json
```diff
@@ -1,66 +1,63 @@
...
+      "gulp-jshint": "^2.0.0",
+      "gulp-mocha": "^2.2.0",
+      "gulp-rename": "^1.2.2",
+      "gulp-sourcemaps": "^1.6.0",
+      "gulp-uglify": "^1.5.1",
+      "istanbul": "^0.4.2",
+      "jshint": "^2.9.1",
+      "jshint-stylish": "^2.1.0",
+      "mocha": "^2.3.4",
+      "must": "^0.13.1",
```
AurelioDeRosa: Why changing of the assertion library and removing karma?

test/helpers.js
```diff
@@ -0,0 +1,49 @@
+'use strict';
+
+var expect = require('must');
+var STANDARD_DELAY = 200; // 200ms between nested calls to the API so things settle
```
AurelioDeRosa: If we add this delay I feel we aren't testing the way our users will employ the library. Even in my projects I run the next async call as soon as the callback returns. Any comment @aendrew?

test/helpers.js
```diff
@@ -0,0 +1,49 @@
...
+
+function assertSuccessful(done, cb) {
```
AurelioDeRosa: Is the aim of these helpers to replace only those tests that have those assertions? Stated in other words, if a callback for a test has:

```
expect(err).not.to.exist();
expect(res).to.exist();
expect(xhr).to.be.an.object();
```

I should use this helper, but if the test has more I should not use it?

test/test.auth.js
```diff
@@ -1,59 +1,96 @@
...
+         var gist = github.getGist('f1c0f84e53aa6b98ec03');
+
+         gist.read(function(err, res, xhr) {
+            try {
+               expect(err).not.to.exist();
+               expect(res).to.exist();
+               expect(xhr).to.be.an.object();
+
+               done();
+            } catch(e) {
```
AurelioDeRosa: I'd suggest to avoid names like `e`. They can be confused with `event`.

test/test.auth.js
```diff
@@ -1,59 +1,96 @@
...
+                     done();
 
-      gist.read(function(err, res, xhr) {
-         should.not.exist(err);
-         xhr.should.be.instanceof(XMLHttpRequest);
-         res.should.be.an('object');
+                     return;
+                  }
+               } catch(e2) {
+                  done(e);
```
AurelioDeRosa: I think this line can be avoided an have an empty catch.

test/test.auth.js
```diff
@@ -1,59 +1,96 @@
...
+         });
+
+         user = github.getUser();
       });
 
-      user = github.getUser();
-   });
+      // 200ms between tests so that Github has a chance to settle
+      beforeEach(function(done) {
+         setTimeout(done, 200);
```
AurelioDeRosa: Same comment about the delay I wrote before.

AurelioDeRosa:
In addition to my comments, do you confirm that at this stage the code coverage won't work?

AurelioDeRosa:
I'd also say that we probably should start creating a v.1 branch. This way we can start merging all this work plus other things that will come in that branch, but keep merging the PRs other devs are submitting. Once we complete version 1, we can rebase and complete a massive merge.

gulpfile.js
clayreimann: It does remove sauce completely because for PRs Sauce doesn't work. The intent is to add it back for `master` so we can populate the browser matrix.

package.json
clayreimann: Remove `karma` because we're not doing browser testing any more so `mocha` is sufficient. And changing the assertion library because [both `should` and `mocha` make bad design choices](https://github.com/moll/js-must#asserting-on-property-access)

test/helpers.js
clayreimann: The delay allows the Github's filesystem to settle. Before I added the delay I found that the tests that created/updated files/repos failed nondeterministically because Github's system wasn't always fast enough to reflect how we were setting the tests up.

I suppose we could enable it only for writes if you want.

test/helpers.js
clayreimann: It's to keep the tests dry. We (almost) always asserted 
```
expect(err).not.to.exist();
expect(res).to.exist();
expect(xhr).to.be.an.object();
```
So I captured that here, along with the common logic for handling errors. If you have more assertions that the three above then you simply pass a callback to this helper and it will be executed as well.

test/test.auth.js
clayreimann: Nope, without `done(e)` the test won't fail if something goes wrong in the preceding `try` block

AurelioDeRosa: You have it already in the next line (65). You capture the error, do nothing, and then execute the `done(e)` in line 65.

PR #289, opened 2016-02-17
Organization API support

davidcorbin:
I created two methods that I needed for a project I'm working on. I plan on adding support for the rest of the [Organization API](https://developer.github.com/v3/orgs/) on the weekend.

PR #293, opened 2016-02-24
added createOrgRepo for create organization repos

marcosomma:


AurelioDeRosa:
Hi @marcosomma.

Thank you for your interest in this library. To consider your PR, we need you to write a test for the feature you want to add.

src/github.js
```diff
@@ -315,6 +315,12 @@
          this.createRepo = function (options, cb) {
             _request('POST', '/user/repos', options, cb);
          };
+
+         // Create an Organization repo
+         // -------
+         this.createOrgRepo = function (orgname, options, cb) {
```
AurelioDeRosa: Instead of adding a new parameter, I think we can pass the organization name as part of the options.

src/github.js
```diff
@@ -315,6 +315,12 @@
          this.createRepo = function (options, cb) {
             _request('POST', '/user/repos', options, cb);
          };
+
+         // Create an Organization repo
```
AurelioDeRosa: Overall GitHub has a different set of API for organizations, so we should create this method differently. You can work on top of [this PR](https://github.com/michael/github/pull/289/files) that is already doing this change.

marcosomma:
hi @AurelioDeRosa  thanks! This was just a fast change to make it work. I'll improve the code and also add the test asap I have time :) hopefully during the week end ;)

marcosomma:
Hi @AurelioDeRosa I just finish the tests for this pull request, but I receive this error from the git api:
``` json
{
     "message": "You need admin access to the organization before adding a reposito
ry to it.",
    "error": 403
}
```

As user I used the one present in the ```tests/user.json``` and as Organization ``` openaddresses ```

Could you set this user as admin for this organization? Is the only way to pass my tests.

Please let me know what I have to do :)

Thanks
Marco Somma

marcosomma:
@AurelioDeRosa I fix the problem with the organization :)
I also update the ```test/user.json``` to reuse the organization that I setup for this test user.

In local the ```npm test``` command run 72test and no one fail, but I think that the project have some problem in the ```.travis.yml``` file. I'm not able to pass the Travis build.

Regards
@marcosomma

marcosomma:
I also fix the naming conflict between my PR and [this](https://github.com/michael/github/pull/289/files) one to prevent conflict during the merging.

PR #302, opened 2016-03-10
Use `_requestAllPages` for search endpoints

chrisronline:
This PR changes the all search endpoints to use `_requestAllPages` instead of `_request` to support getting the full set of results. 

My concern is how the flattening process should work - the result set is an array of responses instead of a single response. However, my change does not break tests since it seems the tests do not care about the nature of the response object.

How do you recommend I proceed with this?

Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,12 +22,12 @@ bower install github-api
 
 [![Sauce Test Status](https://saucelabs.com/browser-matrix/githubjs.svg)](https://saucelabs.com/u/githubjs)
 
-**Note**: Starting from version 0.10.8, Github.js supports **Internet Explorer 9**. However, the underlying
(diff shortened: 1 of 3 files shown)


AurelioDeRosa:
Hi.

Thank you for the PR. This is a good change and because it might break quite a few things, I'd rather have it for version 1.0. Any thoughts @clayreimann @aendrew?

clayreimann:
@AurelioDeRosa I don't think this is liable to break code, but if you'd rather hold off then that's fine.

I also think that in general PRs should not bump the package version.

chrisronline:
Removed the package version and changelog additions

clayreimann:
@chrisronline This was just released in v1.2.0

PR #304, opened 2016-03-18
Add repo.collaborators

timwis:
Added the ability to get a list of collaborators and check if a user is a collaborator per [documentation](https://developer.github.com/v3/repos/collaborators/). Note that for the life of me I can't get the tests to pass, even though the code works fine and the tests look like all the others. Figured I'd ask here before diving even further trying to debug the tests.


AurelioDeRosa:
Hi.

This method you've implemented follows (pretty much) an anti-pattern called Boolean trap. Based on a given Boolean value, the method changes its behavior. The behavior should be split in two methods, each with its own semantic. For example:

```js
repo.collaborators()
repo.isCollaborator()
```

src/github.js
```diff
@@ -677,6 +677,15 @@
             });
          };
 
+         // Show repository collaborators
+         // -------
+
+         this.collaborators = function (username, cb) {
+            var command = repoPath + '/collaborators' + (username ? '/' + username : '')
```
AurelioDeRosa: Once you split the method in two there  will be no need for this variable. You can use the string directly into the method call.

timwis:
Thanks for the clarification @AurelioDeRosa - I was attempting to be consistent with the other methods in the library (ie. `user.show`) but I can see what you mean. I've separated out `isCollaborator` and updated the tests and documentation. I also verified that if you pass a falsy variable to `isCollaborator`'s first paramter (`username`) github.com will still throw a 404, which is good.

AurelioDeRosa:
The tests were broken so I had to fix them. Apart from that, everything was good and the PR has been merged.

Issue #320, opened 2016-05-09
Missing method for get all releases


BernhardBezdek:
Getting a list containing all releases for a selected ropository isn't available right now.
It's documented here: 
https://developer.github.com/v3/repos/releases/#list-releases-for-a-repository

clayreimann:
@BernhardBezdek This was just released in v1.2.0

PR #321, opened 2016-05-09
#320 Added a getReleases() Method into Repository entity

BernhardBezdek:
Added the method in this pr.


clayreimann:
👍 Thanks for the contribution!

PR #330, opened 2016-05-14
Fixed Repo.move() and added test

cripplet:
`Repo.move()` previously was broken and did not work as intended; this is a fix for this issue.

- **Deleted private function `Repo._updateTree()`**
- **Updated `Repo.commit()` docs**: as per https://developer.github.com/v3/git/commits/#create-a-commit, arg `tree` is a SHA string previously created (via `Repo.createTree()`), not the tree object itself.
- **Changed `Repo.updateHead()` API**: added `force` arg to match Github API docs (https://developer.github.com/v3/git/refs/#update-a-reference).
- **Removed internal reference to deprecated `Repo.updateTree()`**: previously referenced in `Repo.move()`; this has been addressed.
- **Added `Repo.move()` test**: assert moved file has same contents, assert old path does not exist after update.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## master
 ### Features
+Added Milestone API
+* `Issue.listMilestones`
+* `Issue.getMilestone`
+* `Issue.editMilestone`
+* `Issue.deleteMilestone`
 
 ### Fixes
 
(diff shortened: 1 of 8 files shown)


cripplet:
Should squash the commits when merging.

clayreimann:
Thanks @cripplet. I'll get this out today or tomorrow.

Note to me: this is a breaking change.

clayreimann:
@cripplet This is released in `v2.0` which is building now

Issue #325, opened 2016-05-11
Markdown API


knsh14:
hi.
Is there any plan to create markdown API?
https://developer.github.com/v3/markdown/

clayreimann:
I'm not against it. Feel free to submit a PR

knsh14:
thanks.
I will submit

PR #331, opened 2016-05-16
Markdown API

knsh14:
PR for #325


lib/Markdown.js
```diff
@@ -0,0 +1,39 @@
...
+    * @return {Promise} - the promise for the http request
+    */
+   constructor(auth, apiBase) {
+      super(auth, apiBase);
+   }
+
+   /**
+    * Render html from Markdown text.
+    * @see https://developer.github.com/v3/markdown/#render-an-arbitrary-markdown-document
+    * @param {Object} option
```
clayreimann: should be `options`

PR #332, opened 2016-05-19
Support for /user/repos with visibility and/or affiliation query.

randalpinto:
/user/repos supports "visibility" and "affiliation" parameters which are not compatible with "type". Not adding a default type if the user provides either of these. see: https://developer.github.com/v3/repos/#list-user-repositories

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "github-api",
-  "version": "1.3.1",
+  "version": "1.3.0",
   "license": "BSD-3-Clause-Clear",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "github-api",
-  "version": "1.3.0",
+  "version": "1.3.1",
```
clayreimann: Don't update the version please

clayreimann:
Great, thanks!

PR #333, opened 2016-05-19
Add support for Team creation

mtscout6:
If this looks good to you guys I'll be following up with some team management capabilities as well in another PR.

### DIFF
--- a/lib/Team.js
+++ b/lib/Team.js
@@ -1,0 +1,160 @@
+/**
+ * @file
+ * @copyright  2016 Matt Smith (Development Seed)
+ * @license    Licensed under {@link https://spdx.org/licenses/BSD-3-Clause-Clear.html BSD-3-Clause-Clear}.
+ *             Github.js is freely distributable.
+ */
+
+import Requestable from './Requestable';
+import debug from 'debug';
+const log = debug('github:team');
+
+/**
+ * A Team allows scoping of API requests to a particular Github Organization Team.
+ */
+class Team extends Requestable {
+   /**
+    * Create a Team.
+    * @param {string} [teamId] - the id for the team
+    * @param {Requestable.auth} [auth] - information required to authenticate to Github
+    * @param {string} [apiBase=https://api.github.com] - the base Github API URL
+    */
+   constructor(teamId, auth, apiBase) {
+      super(auth, apiBase);
+      this.__teamId = teamId;
+   }
+
+   /**
+    * Get Team information
+    * @see https://developer.github.com/v3/orgs/teams/#get-team
+    * @param {Requestable.callback} [cb] - will receive the team
+    * @return {Promise} - the promise for the http request
+    */
+   getTeam(cb) {
+      log(`Fetching Team ${this.__teamId}`);
+      return this._request('Get', `/teams/${this.__teamId}`, undefined, cb);
+   }
+
+   /**
+    * List the Team's repositories
+    * @see https://developer.github.com/v3/orgs/teams/#list-team-repos
+    * @param {Requestable.callback} [cb] - will receive the list of repositories
+    * @return {Promise} - the promise for the http request
+    */
+   getRepos(cb) {
+      log(`Fetching repositories for Team ${this.__teamId}`);
+      return this._requestAllPages(`/teams/${this.__teamId}/repos`, undefined, cb);
+   }
+
+   /**
+    * Edit Team information
+    * @see https://developer.github.com/v3/orgs/teams/#edit-team
+    * @param {object} options - Parameters for team edit
+    * @param {string} options.name - The name of the team
+    * @param {string} [options.description] - Team description
+    * @param {string} [options.repo_names] - Repos to add the team to
+    * @param {string} [options.privacy=secret] - The level of privacy the team should have. Can be either one
+    * of: `secret`, or `closed`
+    * @param {Requestable.callback} [cb] - will receive the updated team
+    * @return {Promise} - the promise for the http request
+    */
+   editTeam(options, cb) {
+      log(`Editing Team ${this.__teamId}`);
+      return this._request('PATCH', `/teams/${this.__teamId}`, options, cb);
+   }
+
+   /**
+    * List the users who are members of the Team
+    * @see https://developer.github.com/v3/orgs/teams/#list-team-members
+    * @param {object} options - Parameters for listing team users
+    * @param {string} [options.role=all] - can be one of: `all`, `maintainer`, or `member`
+    * @param {Requestable.callback} [cb] - will receive the list of users
+    * @return {Promise} - the promise for the http request
+    */
+   listMembers(options, cb) {
+      log(`Getting members of Team ${this.__teamId}`);
+      return this._requestAllPages(`/teams/${this.__teamId}/members`, options, cb);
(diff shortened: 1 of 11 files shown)


mtscout6:
What's the best way to go about testing these teams? It seems to me that you are doing everything against an Org that you actually plan to use. Do you have a test organization setup?

mtscout6:
It looks like you are using the [github-api-tests](https://github.com/github-api-tests) org for testing this repo. In order to facilitate some better tests for Teams can you add either:

- add a team "Fixed Test Team" with the test user mikedeboertest
- add me to that Organization so I can set that up myself

mtscout6:
Well that's frustrating, Lint only fails on CI but passes locally.... I'll get through it, just frustrating. Any thoughts on why that is? I'm running `npm run lint` locally.

mtscout6:
Have you guys reduced the number of allowed concurrent builds in Travis? Only one job runs at a time and it's really slow to get through CI if there's issues to resolve.

clayreimann:
I think OSS only gets one concurrent build.

mtscout6:
Note that I added two other dev time features. I changed Travis to use container based infrastructure which should improve travis worker startup by about 20 seconds per job. I also added an ESLint plugin for mocha to ensure things like `it.only` or `describe.only` are not forgotten when committing.

mtscout6:
https://blog.travis-ci.com/2014-07-18-per-repository-concurrency-setting/

I can't see what settings you have applied, but on a Saturday when OSS seems to slow down Travis should be allowing more concurrent jobs. Other OSS projects I've worked on have run concurrently.

clayreimann:
I'll investigate. Don't think we've done anything special.

mtscout6:
Either way I think everything should be wrapped up for this PR unless you guys see anything I may have missed.

clayreimann:
As it happens we did limit concurrent jobs to 1. I've removed that setting

lib/Team.js
```diff
@@ -0,0 +1,172 @@
+/**
+ * @file
+ * @copyright  2013 Michael Aufreiter (Development Seed) and 2016 Yahoo Inc.
```
clayreimann: This probably wants to be you (or whomever you've signed your copyright over to), neither Yahoo nor Michael Aufreiter/Development Seed wrote this code.

lib/Team.js
```diff
@@ -0,0 +1,172 @@
...
+    * Add a member to the Team
+    * @see https://developer.github.com/v3/orgs/teams/#add-team-membership
+    * @param {string} username - can be one of: `all`, `maintainer`, or `member`
+    * @param {object} options - Parameters for adding a team member
+    * @param {string} [options.role=member] - The role that this user should have in the team. Can be one
+    * of: `member`, or `maintainer`
+    * @param {Requestable.callback} [cb] - will receive the membership status of added user
+    * @return {Promise} - the promise for the http request
+    */
+   addUser(username, options, cb) {
```
clayreimann: Do these want to be addMember/removeMember for consistency?

lib/Team.js
```diff
@@ -0,0 +1,172 @@
...
+
+   /**
+    * Get repo management status for team
+    * @see https://developer.github.com/v3/orgs/teams/#remove-team-membership
+    * @param {string} owner - Organization name
+    * @param {string} repo - Repo name
+    * @param {Requestable.callback} [cb] - will receive the membership status of added user
+    * @return {Promise} - the promise for the http request
+    */
+   managesRepo(owner, repo, cb) {
```
clayreimann: `isManagedRepo`?

lib/Team.js
```diff
@@ -0,0 +1,172 @@
...
+    * @see https://developer.github.com/v3/orgs/teams/#add-or-update-team-repository
+    * @param {string} owner - Organization name
+    * @param {string} repo - Repo name
+    * @param {object} options - Parameters for adding or updating repo management for the team
+    * @param {string} [options.permission] - The permission to grant the team on this repository. Can be one
+    * of: `pull`, `push`, or `admin`
+    * @param {Requestable.callback} [cb] - will receive the membership status of added user
+    * @return {Promise} - the promise for the http request
+    */
+   addOrUpdateRepo(owner, repo, options, cb) {
```
clayreimann: `updateRepoManagement`?

lib/Team.js
```diff
@@ -0,0 +1,172 @@
...
+
+   /**
+    * Remove repo management status for team
+    * @see https://developer.github.com/v3/orgs/teams/#remove-team-repository
+    * @param {string} owner - Organization name
+    * @param {string} repo - Repo name
+    * @param {Requestable.callback} [cb] - will receive the membership status of added user
+    * @return {Promise} - the promise for the http request
+    */
+   removeRepo(owner, repo, cb) {
```
clayreimann: `unmanageRepo`?

test/auth.spec.js
```diff
@@ -51,7 +51,7 @@ describe('Github', function() {
                done();
             } catch (e) {
                try {
-                  if (err && err.request.headers['x-ratelimit-remaining'] === '0') {
+                  if (err && err.response.headers['x-ratelimit-remaining'] === '0') {
```
clayreimann: Good catch!

lib/Team.js
mtscout6: Personally I don't see a need for any of that if you declare it for the repo as a whole. I'll change it to my name though.

clayreimann:
I've made suggestions for some different method names, and will try and get a hold of the person who created the `github-api-tests` org to see about adding teams/some committers here as admins.

lib/Team.js
mtscout6: I like that, I'll change it in just a minute.

lib/Team.js
mtscout6: This one is tough, `update` gives a feeling that the repo is already managed, which may not be the case.

mtscout6: I do like adding `Management` to the end of the name though.

mtscout6: For consistency with the `unmanageRepo` name suggestion how does `manageRepo` sound?

lib/Team.js
mtscout6: I just realized that I missed a test for the membership removal, working on that now.

lib/Team.js
mtscout6: :+1:

mtscout6:
I'm going to remove the `removeMembership` method. When trying to add a test for it I keep getting nothing but a `405 Not Allowed` error code. I'm not sure about the root cause of this, so I'm going to pull it out for now.

mtscout6:
Well, I wonder if the membership removal is related to the current spikes happening within GitHub right now: https://status.github.com/

lib/Team.js
clayreimann: Seems like a reasonable choice to me

mtscout6:
I rebased this on the latest changes from master this morning, but for some reason Travis is not getting kicked off...

clayreimann:
@mtscout6 Not to concerned, I assume they pass locally.

clayreimann:
@mtscout6 v2.1 should be going out now

mtscout6:
Thanks!

PR #334, opened 2016-05-19
Added getSingleCommit to Repository

randalpinto:
Added getSingleCommit method to Repository as per https://developer.github.com/v3/repos/commits/#get-a-single-commit

### DIFF
--- a/lib/Repository.js
+++ b/lib/Repository.js
@@ -179,10 +179,10 @@ class Repository extends Requestable {
      * @param {Requestable.callback} cb - will receive the commit information
      * @return {Promise} - the promise for the http request
      */
-    getSingleCommit(ref, cb) {
-       ref = ref || '';
-       return this._request('GET', `/repos/${this.__fullname}/commits/${ref}`, null, cb);
-    }
+   getSingleCommit(ref, cb) {
+      ref = ref || '';
+      return this._request('GET', `/repos/${this.__fullname}/commits/${ref}`, null, cb);
+   }
 
    /**
     * Get tha sha for a particular object in the repository. This is a convenience function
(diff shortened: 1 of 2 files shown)


lib/Repository.js
```diff
@@ -172,6 +172,19 @@ class Repository extends Requestable {
       return this._request('GET', `/repos/${this.__fullname}/commits`, options, cb);
    }
 
+    /**
+     * Gets a single commit information for a repository
+     * @see https://developer.github.com/v3/repos/commits/#get-a-single-commit
+     * @param {string} ref - the reference for the commit, if not provided this method behaves as listCommits
+     * @param {string} ref - the reference for the commit
```
clayreimann: duplicate tag here

lib/Repository.js
```diff
@@ -172,6 +172,19 @@ class Repository extends Requestable {
...
+     * @param {Requestable.callback} cb - will receive the commit information
+     * @return {Promise} - the promise for the http request
+     */
+    getSingleCommit(ref, cb) {
+       ref = ref || '';
```
clayreimann: I would prefer that we fail here if the ref isn't provided to avoid callers getting unexpected returns (`getSingleCommit` sounds like I'll be expecting to receive an object, not an array)

lib/Repository.js
```diff
@@ -172,6 +172,19 @@ class Repository extends Requestable {
       return this._request('GET', `/repos/${this.__fullname}/commits`, options, cb);
    }
 
+    /**
+     * Gets a single commit information for a repository
+     * @see https://developer.github.com/v3/repos/commits/#get-a-single-commit
+     * @param {string} ref - the reference for the commit, if not provided this method behaves as listCommits
```
clayreimann: perhaps use the git term `commit-ish` here instead of `commit`

lib/Repository.js
randalpinto: It does return an object and it does fail. The replacement of ref with '' is just to make sure that if someone calls a branch 'undefined' or 'null' the API might return results when ref wasn't provided. See unit tests showing that it does fail.

PR #336, opened 2016-05-20
Create travishook.md

Jack-Collins:


clayreimann:
Thanks for the PR, it's a good start but I'm pretty sure your `try`/`catch` block isn't an effective method of handling errors. 

There shouldn't be anything wrong with the call itself–you're probably trying to handle errors that are thrown by GitHub's api. Those errors will be thrown asynchronously (and since you pass a callback they won't even actually be `throw`n, they'll be passed into the callback.

PR #342, opened 2016-05-31
Added getReadme to Repository

mversluys:
Derived from the very similar getContents. Provides support for the
preferred method of retrieving the README from a repository.

### DIFF
--- a/lib/Repository.js
+++ b/lib/Repository.js
@@ -405,6 +405,20 @@ class Repository extends Requestable {
    }
 
    /**
+    * Get the README of a repository
+    * @see https://developer.github.com/v3/repos/contents/#get-the-readme
+    * @param {string} ref - the ref to check
+    * @param {boolean} raw - `true` if the results should be returned raw instead of GitHub's normalized format
+    * @param {Function} cb - will receive the fetched data
+    * @return {Promise} - the promise for the http request
+    */
+   getReadme(ref, raw, cb) {
+      return this._request('GET', `/repos/${this.__fullname}/readme`, {
+         ref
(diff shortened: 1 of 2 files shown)


lib/Repository.js
```diff
@@ -405,6 +405,20 @@ class Repository extends Requestable {
...
    /**
+    * Get the README of a repository
+    * @see https://developer.github.com/v3/repos/contents/#get-the-readme
+    * @param {string} ref - the ref to check
+    * @param {boolean} raw - `true` if the results should be returned raw instead of GitHub's normalized format
+    * @param {Function} cb - will receive the fetched data
+    * @return {Promise} - the promise for the http request
+    */
+   getReadme(ref, raw, cb) {
+      return this._request('GET', `/repos/${this.__fullname}/readme`, {
```
clayreimann: One line please

mversluys: Was attempting the match the format of existing functions like getContents
https://github.com/michael/github/blob/master/lib/Repository.js#L400

Did you mean that it should look like this?

```javascript
getReadme(ref, raw, cb) {
  return this._request('GET', `/repos/${this.__fullname}/readme`, { ref }, cb, raw)
}
```

clayreimann: Thanks for matching the existing style (if only that was what I meant for it to be, stupid auto-formatters 😜), I'll clean both of them up. They'll both be:

```javascript
get*(ref, raw, cb) {
  return this._request('GET', `/repos/${this.__fullname}/readme`, {ref}, cb, raw);
}
```

PR #347, opened 2016-06-10
JSDoc: remove "foo" parameter

kribblo:


kribblo:
"All checks have failed" when changing a comment (jsdoc), really?

clayreimann:
@kribblo Yeah, the github API can be flaky–especially in unit-testing situations where we can be making filesystem changes which don't necessarily settle in the ~50ms we allow. If you add `[skip ci]` to your commit message the you can avoid the whole mess.

kribblo:
> If you add [skip ci] to your commit message the you can avoid the whole mess.

Good tip! (when appropriate)

PR #354, opened 2016-06-18
Rate limit errors

### LINKED ISSUE #348: Error callback not firing when hitting rate limit
I'm using github-api in Node v4.4.5, and I've found that if I exhaust the 60 request/hour unauthenticated rate limit, then certain functions never call their callback. For example, the following script works normally until I hit the limit, then after that the callback never gets called.

```
#!/usr/bin/env node
"use strict";
var GitHub = require("github-api");
var gh = new GitHub({});

var username = "divergentdave";
var user = gh.getUser(username);
user.listRepos(function(error, result, request) {
    console.log("hit callback");
    if (error) {
        console.error(error);
        return;
    }
});

```

```
pi@raspberrypi ~/github-user-branches $ node check-rate-limit.js 
Limit remaining: 1
Reset date: Mon Jun 13 2016 20:45:09 GMT-0500 (CDT)
pi@raspberrypi ~/github-user-branches $ node example.js 
hit callback
pi@raspberrypi ~/github-user-branches $ node example.js 
pi@raspberrypi ~/github-user-branches $ 
```

Other methods, such as `User.getProfile()` properly call the callback with an error object, complete with details of the 403 response from the API server. I haven't arrived at the root cause yet, but it is most likely in `Requestable._requestAllPages`.

I'd be up for working on a PR for this, but first, testing for this issue should probably be done with fixture data for the HTTP responses, rather than hitting the GitHub servers directly. Do you have any preferences for which mocking library to use? I see that [nock](https://www.npmjs.com/package/nock) is a popular choice.

divergentdave:
This fixes #348. So far I have added fixture data and failing tests.


divergentdave:
I think I have fixed the underlying problem in ad6be66; thanks to @clayreimann for the [spot-on analysis](https://github.com/michael/github/issues/348#issuecomment-226791444). This has exposed a problem with the search tests, [which are now failing](https://travis-ci.org/michael/github/jobs/138742854). The Search API [has its own rate limit](https://developer.github.com/v3/search/#rate-limit), and I think the 403 errors are probably due to the four build jobs exhausting the limit. I'm going to mock out the search API too, with pre-recorded responses, to take care of this.

(I also noticed the issue milestone tests fail intermittently. This appears to be a race condition between the four test runners creating and deleting the same milestone.)

package.json
```diff
@@ -49,6 +49,7 @@
     "debug": "^2.2.0",
     "es6-promise": "^3.0.2",
     "js-base64": "^2.1.9",
+    "nock": "^8.0.0",
```
AurelioDeRosa: This is used only in the test files, so it should be a devDependency not a dependency.

divergentdave: :+1: fixed

lib/Requestable.js
```diff
@@ -250,7 +250,13 @@ class Requestable {
 
             response.data = results;
             return response;
-         }).catch(callbackErrorOrThrow(cb, path));
```
clayreimann: Why are you changing this? We should instead make the `callbackErrorOrThrow` function safe to use in this case to keep our error handling the same across pages/non-paged requests.

lib/Requestable.js
```diff
@@ -277,7 +283,9 @@ function getNextPage(linksHeader = '') {
 
 function callbackErrorOrThrow(cb, path) {
    return function handler(response) {
```
clayreimann: This can be made safer by defining it as such: `return function handler({status, statusText, config:{method, url}})`. Though you should double check that the `config:{method, url}` part doesn't give "cannot read property `method` of `undefined`"

lib/Requestable.js
```diff
@@ -250,13 +250,7 @@ class Requestable {
...
             response.data = results;
             return response;
-         }).catch(function(error) {
-            if (cb) {
-               cb(error);
-            } else {
-               throw error;
-            }
-         });
+         }).catch(callbackErrorOrThrow(cb, path));
```
clayreimann: Looking at this again, I think you were on to something with this change in the first place, I think that perhaps it needed to be a bit more nuanced. Maybe something like:

```javascript
.catch(function(error) {
  // we haven't processed this error yet
  if (!(error instanceof ResponseError)) {
    callbackErrorOrThrow(cb, path)(error);

  // we have already processed this error
  } else if (cb) {
    cb(error);
  } else {
    throw error
  }
});
```

Thoughts?

clayreimann:
@divergentdave thanks for the PR, should the tests be passing?

I also looked at your change to the `catch` on `_requestAllPages` and added a comment on 3c70db0, you may have been right the first time 'round.

divergentdave:
Not yet, this still needs some work

divergentdave:
Okay, 9ad89531dabe194494ec3f2957a311b284a094e5 passes tests locally for me. (though some tests are flaky) I'm working on mocking out the search API requests, as described above, but I think the axios and nock libraries might be interfering with each other.

divergentdave:
This is ready to go now. I figured out the search API fixtures, and I squashed some commits to clean up the history. Let me know if you have any questions!

lib/Requestable.js
```diff
@@ -276,10 +276,17 @@ function getNextPage(linksHeader = '') {
...
-   return function handler(response) {
-      let message = `error making request ${response.config.method} ${response.config.url}`;
-      let error = new ResponseError(message, path, response);
-      log(`${message} ${JSON.stringify(response.data)}`);
+   return function handler(object) {
+      let error;
+      if (object.hasOwnProperty('config')) {
+         const {status, statusText, config: {method, url}} = object;
+         let message = (`${status} error making request ${method} ` +
+                        `${url}: "${statusText}"`);
```
clayreimann: Is there a reason to break this into two lines?

divergentdave: Not anymore, I think the expressions were longer at the time. Fixed

clayreimann:
@divergentdave Thanks for the great PR!

divergentdave:
You're welcome, happy to contribute!

AurelioDeRosa:
Thank you very much indeed!

PR #355, opened 2016-06-20
refactor: Repository.move filters nodes instead of deleting their sha

AurelioDeRosa:
closes #215

### DIFF
--- a/lib/Repository.js
+++ b/lib/Repository.js
@@ -588,11 +588,9 @@ class Repository extends Requestable {
                if (ref.path === oldPath) {
                   ref.path = newPath;
                }
-               if (ref.type === 'tree') {
-                  delete ref.sha;
-               }
                return ref;
-            });
+            })
+            .filter((ref) => ref.type !== 'tree');
             return this.createTree(newTree);
          })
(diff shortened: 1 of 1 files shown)


AurelioDeRosa:
@clayreimann do you know why the build keeps failing?

clayreimann:
They need better isolation. The tests are pretty fragile, if you look at the build logs for each build that fails they all fail in different ways. Perhaps we just need to slow them down further (maybe 300-400ms between calls), but I think the thing we can do for the tests is to make sure that each test can run independently–so remove cross-test dependencies–and set mocha to retry failed tests a couple times. I've often found that running a build a second time will result in the build passing.

Also, we should update `travis.yml` to only build master and the tags as in [this example](https://github.com/yahoo/elide/blob/master/.travis.yml#L17-L20)

lib/Repository.js
```diff
@@ -588,11 +588,9 @@ class Repository extends Requestable {
                if (ref.path === oldPath) {
                   ref.path = newPath;
                }
-               if (ref.type === 'tree') {
-                  delete ref.sha;
-               }
                return ref;
-            });
+            })
+            .filter((ref) => ref.type !== 'tree');
```
clayreimann: Perhaps add whitespace for clarity.

clayreimann: between this line and the return

PR #356, opened 2016-06-27
Fixing missing protocol in CDN example in README

mainiak:
Fixing missing protocol in CDN example in README


AurelioDeRosa:
I think it was made on purpose but I do agree that it's an anti-pattern now. If the HTTPS version is available, it should be used.

PR #359, opened 2016-07-03
chore: Improved description to access the documentation

AurelioDeRosa:


KristerV:
It's like a 10% improvement. When I'm searching for a library I don't read through all the text provided, I just need to get an idea of what the thing is.

If possible I'd suggest having the most recent version in bold with a complementary statement:
- **2.3.0 most recent documentation**

I know it's not quite accurate, but that would definitely get the point across.

AurelioDeRosa:
The title of the section already read "API documentation", I thought that was enough to describe the purpose.

clayreimann:
I've got some ideas for how to improve this. I'll open a PR today or tomorrow.

AurelioDeRosa:
Merging it for now. We can always improve it later on, but I haven't seen much traction on this issue in the latest months.

PR #360, opened 2016-07-03
feature(watching): Add support for ignoring or subscribing to repos

mtscout6:
See: https://developer.github.com/v3/activity/watching/


mtscout6:
ping

lib/Watching.js
```diff
@@ -0,0 +1,95 @@
...
+
+import Requestable from './Requestable';
+import debug from 'debug';
+const log = debug('github:watching');
+
+/**
+ * Watching a Repository registers the user to receive notifications on new discussions, as well as events in the user's
+ * activity feed.
+ */
+export default class Watching extends Requestable {
```
AurelioDeRosa: At the moment we use `module.exports` for exporting classes. I'm not sure about the reason as this was made by @clayreimann but I think it's because we aren't using the babel plugin to revert the export of defaults to the behavior of Babel 5.

lib/Watching.js
```diff
@@ -0,0 +1,95 @@
...
+   /**
+    * Subscribe to repo
+    * @see https://developer.github.com/v3/activity/watching/#set-a-repository-subscription
+    * @param {string} owner - username or organization
+    * @param {string} repo - repo name
+    * @param {Requestable.callback} [cb] - subscription status
+    * @return {Promise} - the promise for the http request
+    */
+   subscribe(owner, repo, cb) {
+      log(`Subscribing to ${owner}/${repo}`);
```
AurelioDeRosa: I think we can remove this and the other logs. Possibly we might want to use them only for weird situations. Any opinion @clayreimann?

lib/Watching.js
```diff
@@ -0,0 +1,95 @@
...
+      return this._request204or404(`/repos/${owner}/${repo}/subscription`, undefined, cb, 'DELETE');
+   }
+
+   /**
+    * General subscription method
+    * @private
+    * @see https://developer.github.com/v3/activity/watching/#set-a-repository-subscription
+    * @param {string} owner - username or organization
+    * @param {string} repo - repo name
+    * @param {bool} subscribed - subscribed or ignored
```
AurelioDeRosa: This should be `boolean`.

lib/Watching.js
clayreimann: the logs only run if they are enabled, the idea with including them is that when we need extra information we have the tooling in place to help track down an end-user reported issue. (Logs are enabled via the `DEBUG` command-line argument or flipping the switch in your browser)

lib/Watching.js
clayreimann: I didn't know there was such a plugin. I forget the exact reason that we're using `module.exports` but it boiled down to nothing works if you use the `export default` syntax.

If there's a plugin that fixes things while allowing us to use the new syntax then I'm all for it.

lib/Watching.js
AurelioDeRosa: What I meant was, is there any sense in this specific case in stating the obvious. For example you call `something.subscribe()` and you read on the console `"You have called the subscribe method"`.

lib/Watching.js
AurelioDeRosa: I use [babel-plugin-add-module-exports](https://www.npmjs.com/package/babel-plugin-add-module-exports).

mtscout6: I've used that plugin before as well. I'll get it wired up.

AurelioDeRosa: I suggest to do it in a separate PR.

lib/Watching.js
mtscout6: Since they don't get output unless you explicitly turn them on I don't see the harm here. This can be useful for diagnostic purposes when needed. There is logging of this sort in the base class, however if you enable it there you enable it for everything in this library. Having these here will allow you to only turn on logging in this section.

lib/Watching.js
mtscout6: The commit is already in, it was a two line change and doesn't require me to change this. I can do that if you like though.

mtscout6: If I do a separate PR for this, then I'll tackle the babel-runtime issue as well. I'll crank that out real quick.

mtscout6: Alright it is in PR #366

PR #365, opened 2016-07-14
chore: Fixed a typo in the name of the updatePullRequest method

### LINKED ISSUE #362: typo updatePullRequst vs updatePullRequest
seems to be a typo in the method name

http://michael.github.io/github/docs/2.3.0/Repository.html#updatePullRequst

AurelioDeRosa:
fixes #362

### DIFF
--- a/lib/Repository.js
+++ b/lib/Repository.js
@@ -491,7 +491,7 @@ class Repository extends Requestable {
     * @param {Requestable.callback} [cb] - will receive the pull request information
     * @return {Promise} - the promise for the http request
     */
-   updatePullRequst(number, options, cb) {
+   updatePullRequest(number, options, cb) {
       return this._request('PATCH', `/repos/${this.__fullname}/pulls/${number}`, options, cb);
    }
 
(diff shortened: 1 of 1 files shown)


lib/Repository.js
```diff
@@ -491,7 +491,7 @@ class Repository extends Requestable {
     * @param {Requestable.callback} [cb] - will receive the pull request information
     * @return {Promise} - the promise for the http request
     */
-   updatePullRequst(number, options, cb) {
+   updatePullRequest(number, options, cb) {
```
clayreimann: I'm guessing that you'll need to update some tests because of this. (Also this is a breaking change)

AurelioDeRosa: I don't think we have tests for this. This was added a few days back.

I agree this is BC but it seemed too much to need to do a bump of a major version for a typo.

mtscout6: If you do an alias add a log that will be displayed regardless of the debug preferences that this is deprecated so you can remove it in subsequent major releases.

AurelioDeRosa: This has been  open for a while now. I'm going to add an alias and add a warning as suggested.

PR #372, opened 2016-08-26
add getBranch to repository

deecewan:
This method is present on Github, and provides a lot more detail about a particular branch than `listBranches`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ docs/
 dist/
 coverage/
 node_modules/
+.idea/
 
 .DS_Store
 npm-debug.log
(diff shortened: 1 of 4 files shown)


clayreimann:
Did this not do what you had hoped, or did you have other issues?

deecewan:
well, it did what I wanted it to.  but the rest of my commits were trying to get the thing to build if i installed into NPM from git.  There is some magical build proc in the background.

If we're honest, I didn't know that any extra commits I made would appear on here.  So I closed.  The only commit needed is 86ee93a

AurelioDeRosa:
Removed the line from the `.gitignore` file and changed the commit message. PR merged.

PR #376, opened 2016-09-02
Add support for deply key APIs

mironal:
add this apis. https://developer.github.com/v3/repos/keys/

### DIFF
--- a/lib/Repository.js
+++ b/lib/Repository.js
@@ -551,6 +551,49 @@ class Repository extends Requestable {
    }
 
    /**
+    * List the deploy keys for the repository
+    * @see https://developer.github.com/v3/repos/keys/#list-deploy-keys
+    * @param {Requestable.callback} cb - will receive the list of hooks
+    * @return {Promise} - the promise for the http request
+    */
+   listKeys(cb) {
+      return this._request('GET', `/repos/${this.__fullname}/keys`, null, cb);
+   }
+
+   /**
+    * Get a deploy key for the repository
+    * @see https://developer.github.com/v3/repos/keys/#get-a-deploy-key
+    * @param {number} id - the id of the webook
+    * @param {Requestable.callback} cb - will receive the details of the webook
+    * @return {Promise} - the promise for the http request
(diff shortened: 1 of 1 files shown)


lib/Repository.js
```diff
@@ -551,6 +551,49 @@ class Repository extends Requestable {
...
+    * @return {Promise} - the promise for the http request
+    */
+   listKeys(cb) {
+      return this._request('GET', `/repos/${this.__fullname}/keys`, null, cb);
+   }
+
+   /**
+    * Get a deploy key for the repository
+    * @see https://developer.github.com/v3/repos/keys/#get-a-deploy-key
+    * @param {number} id - the id of the webook
```
AurelioDeRosa: I'd say `{number|string}`

lib/Repository.js
```diff
@@ -551,6 +551,49 @@ class Repository extends Requestable {
...
+    * @return {Promise} - the promise for the http request
+    */
+   createKey(options, cb) {
+      return this._request('POST', `/repos/${this.__fullname}/keys`, options, cb);
+   }
+
+   /**
+    * Delete a deploy key
+    * @see https://developer.github.com/v3/repos/keys/#remove-a-deploy-key
+    * @param {number} id - the id of the webhook to be deleted
```
AurelioDeRosa: Same as the previous comment.

AurelioDeRosa:
Apart from the two very minor comments, the PR looks good.

lib/Repository.js
mironal: I think `{number|string}` is better. but I think that "united feeling" will be lost with the fix only here. Because `id` is also in other places is `{number}`. For example, `getHook`. 

only here or fix all?

clayreimann: I would leave it as `number`, because the objects' ids are numbers in github's API (for most objects).

```javascript
"owner": {
      "login": "octocat",
      "id": 1,
```

mironal:
For now it was fixed comments.

PR #383, opened 2016-09-18
Change array response data type check

ngs:
Hi,

I tested this library using [jest] + [nock], but I got an error with error message:

```
Error: cannot figure out how to append [object Object] to the result set
        at ResponseError.Error (native)
        at new ResponseError (/Users/ngs/.go/src/github.com/ngs/wiplock/node_modules/github-api/dist/components/Requestable.js:103:92)
        at /Users/ngs/.go/src/github.com/ngs/wiplock/node_modules/github-api/dist/components/Requestable.js:276:25
        at run (/Users/ngs/.go/src/github.com/ngs/wiplock/node_modules/core-js/modules/es6.promise.js:87:22)
        at /Users/ngs/.go/src/github.com/ngs/wiplock/node_modules/core-js/modules/es6.promise.js:100:28
        at flush (/Users/ngs/.go/src/github.com/ngs/wiplock/node_modules/core-js/modules/_microtask.js:18:9)
        at process._tickCallback (internal/process/next_tick.js:103:7)
```

and I found this comparison was returned `false`

```
response.data instanceof Array
```

https://github.com/michael/github/blob/master/lib/Requestable.js#L231

So I changed this to check the `data` has `push` method.

[jest]: https://facebook.github.io/jest/
[nock]: https://github.com/node-nock/nock

### DIFF
--- a/lib/Requestable.js
+++ b/lib/Requestable.js
@@ -228,9 +228,9 @@ class Requestable {
       return this._request('GET', path, options)
          .then((response) => {
             let thisGroup;
-            if (response.data instanceof Array) {
+            if (response.data && typeof response.data.push === 'function') {
                thisGroup = response.data;
-            } else if (response.data.items instanceof Array) {
+            } else if (response.data.items && typeof response.data.items.push === 'function') {
                thisGroup = response.data.items;
             } else {
                let message = `cannot figure out how to append ${response.data} to the result set`;


AurelioDeRosa:
Checking if an object has a `push()` function doesn't mean it's an array. So, in my opinion this PR doesn't improve the situation.

ngs:
Do you have any idea to solve this problem when `Array` constructor is different between 2 objects while testing?

AurelioDeRosa:
Before finding  a solution, we need to understand what the problem is really.  The test gives you `false` under what circumstances? What API were you using? Did you expect to have an array or an object (as it's the case in your call)?

ngs:
Here is the example code that reproduces the error

- https://github.com/ngs/github-api-nock-example
- CI result is https://circleci.com/gh/ngs/github-api-nock-example/2

You can check this patch fixes this from

- Diff: https://github.com/ngs/github-api-nock-example/commit/a66ccabc76ac0e0636f08214bece7c3febffc7c5
- CI: https://circleci.com/gh/ngs/github-api-nock-example/4

AurelioDeRosa:
The problem is that you're mocking the data returned incorrectly. The returned results should returned into a `data` property. I'm closing this PR as the issue has nothing to do with the library.

PR #384, opened 2016-09-22
Add Projects api and remove lib from NPM package

klarkc:
Adds the new [Projects API](https://developer.github.com/v3/repos/projects/)

The tests are passing, but the lint contains errors from master.

The lib folder are not used in the NPM package, so I put in the .npmignore

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,6 +1,6 @@
 docs/
 coverage/
 node_modules/
-
+lib/
 .DS_Store
 sauce.json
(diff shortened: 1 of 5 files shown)


klarkc:
I found that the test repo was causing a lot of trouble in tests, so I wrote the clearRepo function that basically clear the test repo. But because of this change, some other tests that not create testing repos now are broken. I am working on these tests to first create the testing repo and after that do the tests.

klarkc:
I had to do the same thing with teams, so I wrote also the `clearTeams` function.

klarkc:
@clayreimann The travis is testing with [concurrency 1](https://blog.travis-ci.com/2014-07-18-per-repository-concurrency-setting/)? As the tests are made in only one user and organization, the running of parallel tests can cause some unknown errors. Just checking.

Issue #241, opened 2015-10-21
Add a TypeScript definition


aendrew:
This will help adoption by people who use TypeScript and doesn't impact the library any other way.

I'll do this because I want it and have experience with TS.

aendrew:
Anyone interested, I've started work at aendrew/github#typescript_def. It's completely unfinished and I'm struggling with the syntax, so any help is quite welcome.

aendrew:
Apparently you can generate these via jsDoc. I'd really prefer to write these all in one place and then have tooling (Possibly [this](https://github.com/fivetran/typescript-closure-tools)) generate any copies, so will add these to the definition for the moment and then migrate that documentation into the actual code itself sometime around 0.11.0 (I'd do so now but it'd change the shape of the code enough that it'd likely cause a large number of existent PRs to conflict).

AurelioDeRosa:
@aendrew Are you still interested in this?

aendrew:
@AurelioDeRosa Yep, absolutely, just a bit delayed. Probably a good thing though; ambient definitions have changed a bit since I posted this. Currently researching the best way to add a def to a lib.

clayreimann:
@aendrew We're all jsdoc'd up so integrating this shouldn't be too hard if you want to.

smakinson:
@aendrew Have you made any progress on your research?

aendrew:
Here's a first pass, I have no idea if it actually works yet — I've merely generated the d.ts from jsDoc and haven't had time to play with the definition beyond that: https://github.com/aendrew/github/tree/d.ts

aendrew:
See above PR if you want to test the definition I've been working on. Automatic generation is about 95% of the way there; once tsd-jsdoc is fixed upstream it'll be worth adding d.ts generation as a build step.

PR #386, opened 2016-10-04
[WIP] 🌟 New: Adds TypeScript definition generation

aendrew:
As per https://github.com/michael/github/issues/241#issuecomment-250444749 I've been working on a TypeScript definition.

The d.ts included in this PR is 95% automated, 5% by hand. `tsd-jsdoc` currently has a [bug](https://github.com/englercj/tsd-jsdoc/issues/6) that moves types into class definitions, where they fail. They need to be in a namespace as per the file included herein.

I'm leaving this as WIP for the moment so people can troubleshoot the issue with tsd-jsdoc; the goal is to have the d.ts file generated on every PR so it remains constant as the API all changes.

### DIFF
--- a/lib/Requestable.js
+++ b/lib/Requestable.js
@@ -21,7 +21,7 @@ if (typeof Promise === 'undefined') {
  */
 class ResponseError extends Error {
    /**
-    * Construct a new ResponseError
+    * @typedef {ResponseError} Requestable.Error
     * @param {string} message - an message to return instead of the the default error message
     * @param {string} path - the requested path
     * @param {Object} response - the object returned by Axios
@@ -37,14 +37,15 @@ class ResponseError extends Error {
 
 /**
  * Requestable wraps the logic for making http requests to the API
+ * @class Requestable
  */
 class Requestable {
    /**
     * Either a username and password or an oauth token for Github
-    * @typedef {Object} Requestable.auth
-    * @prop {string} [username] - the Github username
-    * @prop {string} [password] - the user's password
-    * @prop {token} [token] - an OAuth token
+    * @typedef Requestable.auth
+    * @property {string} [username] - the Github username
+    * @property {string} [password] - the user's password
+    * @property {string} [token] - an OAuth token
     */
    /**
     * Initialize the http internals.
@@ -121,7 +122,7 @@ class Requestable {
 
    /**
     * if a `Date` is passed to this function it will be converted to an ISO string
-    * @param {*} date - the object to attempt to cooerce into an ISO date string
+    * @param {Date} date - the object to attempt to cooerce into an ISO date string
     * @return {string} - the ISO representation of `date` or whatever was passed in if it was not a date
     */
    _dateToISO(date) {
@@ -136,19 +137,19 @@ class Requestable {
     * A function that receives the result of the API request.
     * @callback Requestable.callback
     * @param {Requestable.Error} error - the error returned by the API or `null`
-    * @param {(Object|true)} result - the data returned by the API or `true` if the API returns `204 No Content`
-    * @param {Object} request - the raw {@linkcode https://github.com/mzabriskie/axios#response-schema Response}
+    * @param {axios.Response | boolean} result - the data returned by the API or `true` if the API returns `204 No Content`
+    * @param {axios.RequestOptions} request - the raw {@linkcode https://github.com/mzabriskie/axios#response-schema Response}
     */
    /**
     * Make a request.
     * @param {string} method - the method for the request (GET, PUT, POST, DELETE)
     * @param {string} path - the path for the request
-    * @param {*} [data] - the data to send to the server. For HTTP methods that don't have a body the data
+    * @param {any} [data] - the data to send to the server. For HTTP methods that don't have a body the data
     *                   will be sent as query parameters
     * @param {Requestable.callback} [cb] - the callback for the request
     * @param {boolean} [raw=false] - if the request should be sent as raw. If this is a falsy value then the
     *                              request will be made as JSON
-    * @return {Promise} - the Promise for the http request
+    * @return {axios.Promise} - the Promise for the http request
     */
    _request(method, path, data, cb, raw) {
       const url = this.__getURL(path);
@@ -188,7 +189,7 @@ class Requestable {
     * @param {Object} data - any query parameters for the request
     * @param {Requestable.callback} cb - the callback that will receive `true` or `false`
     * @param {method} [method=GET] - HTTP Method to use
-    * @return {Promise} - the promise for the http request
+    * @return {axios.Promise} - the promise for the http request
     */
    _request204or404(path, data, cb, method = 'GET') {
       return this._request(method, path, data)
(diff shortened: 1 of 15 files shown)


pascalwhoop:
I like the PR but don't quiet understand why all the JSDoc needs to be touched to add typing support. Also when I copy the files into my node_modules folder (manually since its not yet in the repo), the axios. namespace gives me errors all over the place. How did it look like for you @aendrew ?

mohsen1:
I looked at this but it's a lot more work that I expected. I resolved the merge conflict anyways. 

@aendrew please merge my PR to your fork so if someone wants to pick it up they can: https://github.com/aendrew/github/pull/1

aendrew:
Sorry for the really slow response to this issue!

@pascalwhoop I'm using JSDoc to generate the TypeScript definitions, and Axios returns a different Promise interface from the standard NodeJS promise. There are a few others that have been changed to allow tsd-jsdoc to properly generate interfaces.

@mohsen1 Cheers, thanks for doing that! I'll try to review and merge that sometime soon (ideally tomorrow but have been rather busy as of late).

tsd-jsdoc has released a new version and I plan to pick this up again shortly. Hopefully the new version will resolve some of the issues with interface generation that resulted in me having to tweak the JSDoc more than I'd have liked.

akfish:
I was trying to use the generated .d.ts file directly in my project. It seems that all references to `axios.Promise` that are marked as 'the Promise for the http request' should be `axios.AxiosPromise`, according to [axios/index.d.ts](https://github.com/mzabriskie/axios/blob/4976816808c4e81acad2393c429832afeaf9664d/index.d.ts#L109). 

Those axios related typings are broken since axios@0.15.3, which is the version I got when I installed github-api@3.0.0 with yarn.

lib/Requestable.js
```diff
@@ -136,19 +137,19 @@ class Requestable {
     * A function that receives the result of the API request.
     * @callback Requestable.callback
     * @param {Requestable.Error} error - the error returned by the API or `null`
-    * @param {(Object|true)} result - the data returned by the API or `true` if the API returns `204 No Content`
-    * @param {Object} request - the raw {@linkcode https://github.com/mzabriskie/axios#response-schema Response}
+    * @param {axios.Response | boolean} result - the data returned by the API or `true` if the API returns `204 No Content`
+    * @param {axios.RequestOptions} request - the raw {@linkcode https://github.com/mzabriskie/axios#response-schema Response}
```
akfish: I couldn't find `axios.Response` and `axios.RequestOptions` in axios/index.d.ts.
Also the callback signature doesn't seem to match how it is used [here](https://github.com/github-tools/github/blob/f075749ed4234c7d76f122879523fbca2eb01f86/lib/Requestable.js#L189):
```js
cb(null, response.data, response);
```
Shouldn't it be:
```ts
/**
 * A function that receives the result of the API request.
 * @callback Requestable.callback
 * @param {Requestable.Error} error - the error returned by the API or `null`
 * @param {any | boolean} result - the data returned by the API or `true` if the API returns `204 No Content`
 * @param {axios.AxiosResponse} response - the raw {@linkcode https://github.com/mzabriskie/axios#response-schema Response}
 */
type callback = (
  error: Requestable.Error,
  result: (any | boolean), // https://github.com/mzabriskie/axios/blob/4976816808c4e81acad2393c429832afeaf9664d/index.d.ts#L48
  response: axios.AxiosResponse // Not request but response
) => void;
```

akfish:
Here is my standalone .d.ts file that works with webpack/ts-loader toolchain:
https://gist.github.com/akfish/b9a42945f1e681ed173b0b06365303a7

aendrew:
Given the codebase has moved along quite a lot and there's a new version of `tsd-jsdoc`, I've redone the TypeScript generation and made it so that no code changes are actually necessary. It's not quite automated yet, but the definition in the above commit should work a lot better. See above commit.

The steps needed to generate a `d.ts` file upon release are:

1. Generate d.ts via `npm run make-ts-def`
2. Add `import {AxiosPromise} from 'axios'` at top of newly-created `github-api.d.ts`
3. Replace all instances of `Promise` with `AxiosPromise`
4. Move the `Requestable` class into the `Requestable` module; `export default` it.
5. Replace all instances of `callback` with `Requestable.callback` (except in `Requestable` module itself)
6. Replace all instances of `auth` with `Requestable.auth` (except in `Requestable` module itself)
7. Fix any line that includes nested `options` values, i.e.: `options.author?: any, options.commiter?: any, options.encode?: boolean`
8. Change any `Params` type to `any`
9. Fix errors re: optional arguments (add ?)
10. Change `Requestable.auth` to two overloaded interfaces, ensuring both `Requestable.auth` and `Requestable.callback` are exported.

This admittedly isn't the best TypeScript def; it has far too many `any` types in it. But it's definitely a start, and so far this is the only GitHub API lib around that has anything resembling a TypeScript def.

To improve this in the future:
* Replace `Promise` with `AxiosPromise` in JSDoc. This will help a lot as it will get rid of a lot of the `any` types.
* Monitor [this bug](https://github.com/englercj/tsd-jsdoc/issues/23), which will fix the issue with parsing `options` interfaces
 * Probably worth somehow creating generic interfaces for options and params — having these documented as part of the definition greatly increases its usability.
* Figure out why Requestable isn't exported as part of Requestable `module`

The remaining steps can probably be accomplished using codemods or somesuch.

akfish:
Actually my typing was based on the automatic generation. So I can't take credit for that. :)

I took a look at the typing. It would be easier if some automatic tests are included. Here's how I set it up based on [DefinitelyTyped's method](https://github.com/DefinitelyTyped/DefinitelyTyped#how-can-i-contribute):

I changed the `tsconfig.json` to:
```json
{
    "compilerOptions": {
        "module": "commonjs",
        "lib": [
            "es6"
        ],
        "noImplicitAny": true,
        "noImplicitThis": true,
        "strictNullChecks": false,
        "baseUrl": "./",
        "typeRoots": [
            "./"
        ],
        "types": [],
        "noEmit": true,
        "forceConsistentCasingInFileNames": true
    },
    "files": [
        "github-api.d.ts",
        "tests.ts"
    ]
}
```

And added this line to `package.json`:

```json
{
  // other lines
  "types": "github-api.d.ts",
  // other lines
}
```

And added `tests.ts`:

```ts
// Write tests here
// This file will be compiled by tsc to check typings
// It will not be executed
import GitHub = require('github-api')

let gh = new GitHub()
```

Then you just run `tsc` as the test.

Now here're some issues I spotted:

* I believe you will need this line at the beginning of the generated d.ts file:
  ```ts
  /// <reference types="axios" />
  ```
* Compiler error:
  > tests.ts(3,10): error TS2351: Cannot use 'new' with an expression whose type lacks a call or construct signature.

  This suggests the exported signatures are not correct.
  Adding:
  ```ts
  exports = GitHub
  ```
  seems to fix this issue. But I'm not familiar with the code base so I'm not sure if this is a correct fix.

akfish:
See https://github.com/akfish/github/tree/d.ts

aendrew:
@akfish It might be the `tsconfig.json` that's causing the issue, IIRC setting `types` and/or `typeRoots` will prevent `node_modules/` and thus the Axios defs won't be picked up (from what I understand, using triple-slash directives in defs is considered a bad practice).

Will take a closer look sometime tomorrow!

akfish:
No problem.

`/// <reference types="..." />` is used for declaring type dependencies. It's widely used in `@types` packages. Though a second look at the [document](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html) seems to suggest that it's only for `@types/` packages. I'll verify this later.

akfish:
I just verified that the `/// <reference types="..." />` is not necessary for npm package typings. There's no need to change `tsconfig.json` too. It must be a problem with my global TypeScript version yesterday.

dvargas92495:
What's the status here?

I'm interested in this enhancement. Could I take over if no one's working on it?

aendrew:
@dvargas92495 Feel free! I don't expect I'll ever finish it tbth.

PR #390, opened 2016-10-13
feature(repository): add updateStatus and updateRepository method

nmoya:
Update status GitHub API docs: https://developer.github.com/v3/repos/statuses/
Update repository GitHub API docs: Githup API docs: https://developer.github.com/v3/repos/#edit


nmoya:
Other existing tests are failing, the test that I added is passing:
`  ✓ should update commit status`

PR #397, opened 2016-11-12
Adds missing `getCombinedStatus` method to Repository.js

jamsinclair:
As I was about to use this great library I noticed it was missing the exact method I was using it for.
See https://developer.github.com/v3/repos/statuses/#get-the-combined-status-for-a-specific-ref

So thought I better add it for the next person 😸 

__Changes:__
- Adds `getCombinedStatus` to the Repository.js module
- Adds simple test case for `getCombinedStatus` method
- Also fixes the only the only lint error in Repository.js

Let me know if it needs any changes 👍 

__Edit:__
Added test seems to be passing see:
https://travis-ci.org/michael/github/jobs/175262817#L1148

### DIFF
--- a/lib/Repository.js
+++ b/lib/Repository.js
@@ -232,6 +232,17 @@ class Repository extends Requestable {
    }
 
    /**
+    * Get the combined commit Status for a specific sha, branch or tag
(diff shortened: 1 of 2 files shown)


wking:
I'd like this too.  It looks like the PR needs a rebase?

j-rewerts:
Closing in favour of #580.

Issue #398, opened 2016-11-15
pagination in listIssueComments


bambooom:
I need to fetch all comments (>50) in an issue, but I don't know how to pass params like
```js
{
    per_page: 100,
    page: 2
}
```
to [`listIssueComments`](https://github.com/michael/github/blob/master/lib/Issue.js#L65).

I also find that there is a method called `__requestAllPages`, but only applied to [`listIssues`](https://github.com/michael/github/blob/master/lib/Issue.js#L43).

Is is better to add options params or use `__requestAllPages` in `listIssueComments`? 

Thanks a lot.

PR #399, opened 2016-11-15
add options param to `listIssueComments` to get pagination result

bambooom:
as I found that `_requestAllPages` will be deprecated, so I just add options to `_request` method to solve the problem I met in #398. 😊


metaodi:
To have an options parameter is great, this would make it possible to pass the [`since` parameter](https://developer.github.com/v3/issues/comments/#list-comments-on-an-issue), to only get a subset of all comments.

PR #391, opened 2016-10-14
feat: Issue get/delete/edit/list for labels

vicb:
plus some other minor changes

Issue #392, opened 2016-10-17
Uncaught SyntaxError: Unexpected token import


LarsVoith:
geting this error in Chrome (Uncaught SyntaxError: Unexpected token import) with this line of code 

"import GitHub from 'github-api';"??

clayreimann:
This library is written in ES2015 which, while finalized, is not fully supported in any browser. How are you bringing in the library?

LarsVoith:
I'm using a Cordova/PhoneGap app on ios10 but testing in Safari 10 (Mac) which should support it??
________________________________

clayreimann:
Again, ES2015 is not fully supported, and I'm 95% confident that no-one supports the `import` syntax. If it works in your Cordova app it's because they're pre-processing your sources just like web-devs do with webpack/browserify

alundiak:
Not sure how relevant my comment is going to be here, but if issue still open, and I've recently spotted on the same error, I decided to add my 5 cents here.

First of all, I use `Chrome Canary v63`, where import/export is already supported. Not yet officially in Chrome, but still gives me the touch to the feature (because I don't like web community is so *** slow regarding standards but *** fast regarding enormous number of libraries, frameworks, wrappers, etc.)

I have existed code, which uses import/export, and I was a bit surprised, when I did a few lines of code with `github-tools/github-api`. Before that, I've already tested GitHub API v3 by my own using Fetch API and GitHub API v4 using GraphQL stuff.

So...

**Variant ==A==**
`bower install github-api` => 3.1.0, then added code to my one of my modules. My index.html => `<script type="module" src="./index.js"></script>` and index.js has code `import GitHub from 'bower_components/github-api/lib/GitHub.js'` which gave me errors:
```
GitHub.js:9 GET http://localhost:8080/bower_components/github-api/lib/Gist net::ERR_ABORTED
GitHub.js:10 GET http://localhost:8080/bower_components/github-api/lib/User net::ERR_ABORTED
GitHub.js:11 GET http://localhost:8080/bower_components/github-api/lib/Issue net::ERR_ABORTED
GitHub.js:12 GET http://localhost:8080/bower_components/github-api/lib/Search net::ERR_ABORTED
GitHub.js:13 GET http://localhost:8080/bower_components/github-api/lib/RateLimit net::ERR_ABORTED
GitHub.js:14 GET http://localhost:8080/bower_components/github-api/lib/Repository net::ERR_ABORTED
[...]

PR #401, opened 2016-11-18
refactor tests, update dependencies and Project API

mathieudutour:
- [x] fix lint issue
- [x] update dependencies (:boom: breaking change)
- [x] Fix getContributors method URL (:boom: breaking change) (#393)
- [x] Add getContributorStats method (#393)
- [x] Issue.{get/delete/edit/list} for labels (#391)
- [x] Project API (both org and repo)
- [x] remove Promise polyfill (:boom: breaking change, drop support for node 0.12 without polyfill)
- [x] rename `michael/github` to `github-tool/github`
- [x] add support for preview APIs
- [x] add updateStatus and updateRepository method (#390)
- [x] remove `updatePullRequst` (:boom: breaking change)
- [x] add `user.getEmails`
- [x] refactor tests to not use `TestRepo` (and make them pass)
- [x] add test coverage (https://github.com/github-tools/github/pull/401#issuecomment-261646777)

That's a lot of changes. Tell me if you are interested, I'll clean the history and the package.json


lib/Repository.js
```diff
@@ -496,21 +541,6 @@ class Repository extends Requestable {
...
    /**
     * Update a pull request
-    * @deprecated since version 2.4.0
-    * @see https://developer.github.com/v3/pulls/#update-a-pull-request
-    * @param {number|string} number - the number of the pull request to update
-    * @param {Object} options - the pull request description
-    * @param {Requestable.callback} [cb] - will receive the pull request information
-    * @return {Promise} - the promise for the http request
-    */
-   updatePullRequst(number, options, cb) {
```
clayreimann: Lolz

lib/Requestable.js
```diff
@@ -152,7 +155,13 @@ class Requestable {
     */
    _request(method, path, data, cb, raw) {
       const url = this.__getURL(path);
-      const headers = this.__getRequestHeaders(raw);
+
+      const AcceptHeader = (data || {}).AcceptHeader;
```
clayreimann: `acceptHeader`?

package.json
```diff
@@ -1,92 +1,101 @@
 {
-  "name": "github-api",
-  "version": "2.4.0",
+  "name": "@mathieudutour/github-api",
+  "version": "3.2.0",
```
clayreimann: Is this diff still relevant?

test/team.spec.js
```diff
@@ -56,10 +56,7 @@ describe('Team', function() { // Isolate tests that are based on a fixed team
          .then(function({data: members}) {
             expect(members).to.be.an.array();
 
-            let hasTestUser = members.reduce(
-               (found, member) => member.login === testUser.USERNAME || found,
-               false
-            );
+            const hasTestUser = members.some((member) => member.login === testUser.USERNAME);
```
clayreimann: 👍 Yay conciseness!

release.sh
```diff
@@ -32,5 +32,10 @@ git checkout gh-pages
...
 echo $VERSION >> _data/versions.csv
 git add .
-git co -m "adding docs for v$VERSION"
+git commit -m "adding docs for v$VERSION"
 git push
+
+# switch back to master, build and publish
+git checkout master
+npm run build
+npm publish
```
clayreimann: Isn't this done by CI?

.travis.yml
```diff
@@ -2,22 +2,18 @@ sudo: false
 language: node_js
 node_js:
   - '6'
-  - '5'
-  - '4'
```
mathieudutour: we are not dropping the support for node4, just node 0.12 and earlier. I removed it from the test cause there were some issues with running the test 3 times in parallel. But maybe we should test on node4 only

package.json
mathieudutour: it's not, I will roll back those changes. I wasn't sure you were interested in merging such a PR so I waited for that first ;)

release.sh
mathieudutour: indeed, I'll roll back also

.travis.yml
mtscout6: This may be solved by running them serially: https://docs.travis-ci.com/user/customizing-the-build#Limiting-Concurrent-Builds

README.md
```diff
@@ -2,50 +2,19 @@
...
-* 4.x
-* 0.12
-
-## GitHub Tools
-
-The team behind Github.js has created a whole organization, called [GitHub Tools](https://github.com/github-tools),
-dedicated to GitHub and its API. In the near future this repository could be moved under the GitHub Tools organization
-as well. In the meantime, we recommend you to take a look at other projects of the organization.
-
-## Samples
```
mtscout6: What's the thought behind removing all this documentation? Are you proposing we drop the link to the generated Docs?

README.md
```diff
@@ -62,57 +31,66 @@ gist.create({
...
+   return gist.read();
+}).then(function({data}) {
+   let retrievedGist = data;
+   // do interesting things
 });
 ```
 
 ```javascript
-import GitHub from 'github-api';
+var GitHub = require('github-api');
```
mtscout6: Why downgrade from ECMA2015?

README.md
```diff
@@ -62,57 +31,66 @@ gist.create({
...
 
 // basic auth
-const gh = new GitHub({
+var gh = new GitHub({
```
mtscout6: Again downgrading from ECMA2015?

README.md
```diff
@@ -62,57 +31,66 @@ gist.create({
...
    username: 'FOO',
    password: 'NotFoo'
+   /* also acceptable:
+      token: 'MY_OAUTH_TOKEN'
+    */
 });
 
-const me = gh.getUser();
+var me = gh.getUser(); // no user specified defaults to the user for whom credentials were provided
```
mtscout6: Another downgrade from ECMA2015, I'll stop mentioning all the ones I see now.

lib/Repository.js
mtscout6: @clayreimann You were asking about a breaking change that would necessitate a major version bump. This would be it.

lib/Requestable.js
mathieudutour: the header is call `Accept` so I wasn't sure

test/issue.spec.js
```diff
@@ -2,20 +2,43 @@ import expect from 'must';
...
 import {assertSuccessful} from './helpers/callbacks';
+import getTestRepoName from './helpers/getTestRepoName';
 
 describe('Issue', function() {
    let github;
+   const testRepoName = getTestRepoName();
    let remoteIssues;
 
-   before(function() {
+   before(function(done) {
```
mtscout6: Mocha allows you to return a promise chain instead of passing the done callback.

README.md
mathieudutour: It's not removed, it's rearranged to fit https://github.com/noffle/art-of-readme

test/project.spec.js
```diff
@@ -23,7 +23,7 @@ describe('Project', function() {
       github
         .getUser()
         .createRepo({name: testRepoName})
-        .then(wait())
+        .then(wait(5000))
```
mtscout6: I generally find arbitrary delays like this in tests to be brittle. Perhaps a better solution is to poll for the repo until it comes back successfully.

README.md
mtscout6: My opinion is that ECMA2015 is spec now, so there's no need to downgrade.

lib/Requestable.js
mtscout6: I think @clayreimann is getting at the usage of PascalCasing instead of camelCasing.

mathieudutour: yes I understood

README.md
mathieudutour: sure, but `var` is not deprecated in anyway, so if we can help people understand the repo, I think it's best, regardless of wether it's matching style guide or not, don't you think? (see #392)

test/repository.spec.js
```diff
@@ -349,11 +349,12 @@ describe('Repository', function() {
 
       it('should write to repo', function(done) {
          remoteRepo.writeFile('master', fileName, initialText, initialMessage, assertSuccessful(done, function() {
-            remoteRepo.getContents('master', fileName, 'raw', assertSuccessful(done, function(err, fileText) {
+            wait().then(() => remoteRepo.getContents('master', fileName, 'raw',
```
mtscout6: Why the wait?

mathieudutour: Sometimes the `writeFile` succeed even though github didn't really create the file yet and the `getContents` call fails with 404. `Wait` is there to give time to Github to settle down

README.md
mathieudutour: That might be a solution. Feel free to commit on directly here

test/repository.spec.js
mtscout6: Blind wait's a generally very brittle. Can you pull github for the file and if it doesn't show up within 20 seconds then fail?

mathieudutour: Indeed, I'll see if I can write an easy to use helper. (This can be done in another PR though)

mtscout6: Agreed, I'm fine with it in another PR as well.

mtscout6:
For the changelog generation should we have all these commits squashed with the commit message `chore: fix tests`

clayreimann:
There's nothing here that I really object to, @mtscout6 look good to you too?

mtscout6:
Just the rebase and squash for the change log generation.

mathieudutour:
How is the changelog generation working exactly?

clayreimann:
Changelog generation is manual

mtscout6:
Very well, I thought it was auto generated from the commit messages. Never mind then this is good to go.

Issue #403, opened 2016-11-25
Add support for listing Review Comments on a Pull Request


jedfonner:
Currently it's only possible with this library to list comments on an issue, but those are not the same as "Review Comments" on a PR.

See Github's documentation on [Review Comments](https://developer.github.com/v3/pulls/comments/) vs [Comments](https://developer.github.com/v3/issues/comments/):

> Pull Request Review Comments are comments on a portion of the unified diff. These are separate from Commit Comments (which are applied directly to a commit, outside of the Pull Request view), and Issue Comments (which do not reference a portion of the unified diff).

In order to list review comments on a Pull Request, you need to make a call to
`GET /repos/:owner/:repo/pulls/:number/comments`

versus the call for regular issue comments which is
`GET /repos/:owner/:repo/issues/:number/comments`

kyheng:
GET/rwpos/:owner/:repo/pulls/:number/comments

kyheng:
GET/rwpos/:owner/:repo/issues/:number/comments

PR #404, opened 2016-11-25
Adds method for listing review comments

jedfonner:
For issue #403

I didn't find a great example PR with review comments on which to base the test.  Ideally it could be this one or a new test one.

### DIFF
--- a/test/repository.spec.js
+++ b/test/repository.spec.js
@@ -240,6 +240,16 @@ describe('Repository', function() {
             done();
          }));
       });
+      
+      it('should list review comments', function(done) {
+        const repo = github.getRepo('michael', 'github');
+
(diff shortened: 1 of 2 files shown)


test/repository.spec.js
```diff
@@ -240,6 +240,16 @@ describe('Repository', function() {
             done();
          }));
       });
+      
+      it('should list review comments', function(done) {
+        const repo = github.getRepo('michael', 'github');
```
mtscout6: This project was moved to the github-tools project. You'll want to update this to get the tests working.

mtscout6: You'll also want to update based on the latest changes from master that got the tests working again.

PR #412, opened 2017-01-04
Add methods to fetch gist's commits and specific revision

fkling:
First of all, thank you for this library!

I need to be able to get a specific revision of gist, so I added support for these two API methods:

- https://developer.github.com/v3/gists/#list-gist-commits
- https://developer.github.com/v3/gists/#get-a-specific-revision-of-a-gist

### DIFF
--- a/lib/Gist.js
+++ b/lib/Gist.js
@@ -109,6 +109,27 @@ class Gist extends Requestable {
    }
 
    /**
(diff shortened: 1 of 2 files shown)


clayreimann:
Thanks for the contribution.

fkling:
Thanks for merging so quickly!

PR #418, opened 2017-01-27
Fix incorrect path for delete hook API

kainisoft:
Just came across this error when trying to delete a hook of repos

### DIFF
--- a/lib/Repository.js
+++ b/lib/Repository.js
@@ -603,7 +603,7 @@ class Repository extends Requestable {
     * @return {Promise} - the promise for the http request
     */
    deleteHook(id, cb) {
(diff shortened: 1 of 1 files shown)


lib/Repository.js
```diff
@@ -603,7 +603,7 @@ class Repository extends Requestable {
     * @return {Promise} - the promise for the http request
     */
    deleteHook(id, cb) {
-      return this._request('DELETE', `${this.__fullname}/hooks/${id}`, null, cb);
+      return this._request('DELETE', `/repo/${this.__fullname}/hooks/${id}`, null, cb);
```
clayreimann: do you mean `repos` here?

kainisoft: Yep, you're right. Fixed.

clayreimann:
Thanks for the fix

PR #429, opened 2017-02-26
feature(repository): add getCombinedStatus

### LINKED ISSUE #426: Missing API Method To Get A Commit's Combined Status
I looked all throughout the documentation, and I could not find any method in this library to invoke the GitHub API's [combined status endpoint](https://developer.github.com/v3/repos/statuses/#get-the-combined-status-for-a-specific-ref). It seems that support for it was never added to this library.

The combined status endpoint is **not** the same thing as [listing the statuses for a commit](https://developer.github.com/v3/repos/statuses/#list-statuses-for-a-specific-ref) (I know this library **does** support that). Listing all statuses for a commit shows every single status that has ever been set on the commit specified. In contrast, the combined status endpoint returns a **single** combined status which  indicates only the overall combined status for the commit, taking into account only the most recent status set for each context.

The combined status endpoint allows a single call to the GitHub API to determine whether a specific commit ready to deploy or not, without having to parse through a list of every status on the commit, find the most recent statuses for each context, and then run through the logic manually to determine what the combined state should be based on the specific combination of states from the statuses of each context.

I'm not sure why the combined status endpoint got overlooked when creating this library, as it is not a new addition to the GitHub API (GitHub added it to their API [back in 2014](https://developer.github.com/changes/2014-03-27-combined-status-api/)). The combined status endpoint is a very crucial endpoint, as it is the most reliable way to determine if a particular commit has passed all status checks and is ready to be deployed or not.

I used to use the [Octokat.js](https://github.com/philschatz/octokat.js) library for an [application I'm in the process of developing](https://github.com/compumike08/GitHub_Status_API_GUI), but I didn't like the way Octokat.js handled pagination so I started to switch to this library instead. However when I realized that the combined status endpoint was not supported by this library I had to stop migrating, as I need support for the combined status endpoint in my application.

Is adding support for the [combined status endpoint](https://developer.github.com/v3/repos/statuses/#get-the-combined-status-for-a-specific-ref) to this library something that can be done fairly quickly? I would greatly appreciate it if one this library's developers could give me an estimate of when support for the combined status endpoint could be added and released, as depending on how long it might take I may have to go back to using Octokat.js. Thanks.

compumike08:
I added the `getCombinedStatus` method in `Repository.js` and an associated unit test to implement the GitHub get combined status API endpoint (https://developer.github.com/v3/repos/statuses/#get-the-combined-status-for-a-specific-ref).

**NOTE:** The one unit test which is failing is not actually a problem with the unit tests themselves, but rather is being caused by the test harnesses inability to properly authenticate to GitHub.

closes #426 

@clayreimann 
@mtscout6 
@AurelioDeRosa 
@alexcanessa

### DIFF
--- a/lib/Repository.js
+++ b/lib/Repository.js
@@ -232,6 +232,17 @@ class Repository extends Requestable {
    }
 
    /**
+    * Get the combined view of commit statuses for a particular sha, branch, or tag
+    * @see https://developer.github.com/v3/repos/statuses/#get-the-combined-status-for-a-specific-ref
+    * @param {string} sha - the sha, branch, or tag to get the combined status for
+    * @param {Requestable.callback} cb - will receive the combined status
+    * @returns {Promise}
+    */
+   getCombinedStatus(sha, cb) {
+      return this._request('GET', `/repos/${this.__fullname}/commits/${sha}/status`, null, cb);
+   }
+
+   /**
     * Get a description of a git tree
     * @see https://developer.github.com/v3/git/trees/#get-a-tree
     * @param {string} treeSHA - the SHA of the tree to fetch
--- a/lib/Repository.js
+++ b/lib/Repository.js
@@ -236,7 +236,7 @@ class Repository extends Requestable {
     * @see https://developer.github.com/v3/repos/statuses/#get-the-combined-status-for-a-specific-ref
     * @param {string} sha - the sha, branch, or tag to get the combined status for
     * @param {Requestable.callback} cb - will receive the combined status
-    * @returns {Promise}
+    * @returns {Promise} - the promise for the http request
     */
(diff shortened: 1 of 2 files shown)


nicoecheza:
Any news on this one?

gvilarino:
I'd really find this super useful if we could have it merged. Thanks in advance! :D

jaredallard:
LGTM :shipit:

sime:
Are the failing tests blocking this PR ?

A quick look suggests the failing tests are not specific to this PR.

bfeldkamp3:
@j-rewerts can you take a look at this one?

j-rewerts:
@bfeldkamp3 Apologies for not seeing this earlier. Closing in favour of #580.

PR #450, opened 2017-05-09
Working with two-factor authentication

ddimitrioglo:
As mentioned in github API:
```
In addition to the Basic Authentication credentials, you must send the user's authentication code (i.e., one-time password) in the X-GitHub-OTP header.
```
And we will be able to use it like this:

```
const github = new GitHub({
    username: 'ddimitrioglo',
    password: 'xxxxxxxxxxxx',
    otp: '888999' // <= One-Time-Password
});

let me = github.getUser();

me.listRepos((err, repos) => {
    console.log(repos);
});
```

### DIFF
--- a/lib/Requestable.js
+++ b/lib/Requestable.js
@@ -54,6 +54,7 @@ class Requestable {
          token: auth.token,
          username: auth.username,
          password: auth.password,
+         otp: auth.otp || null
       };
       this.__AcceptHeader = AcceptHeader || 'v3';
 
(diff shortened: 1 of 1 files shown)


j-rewerts:
@clayreimann @CodyGramlich any idea how we would run tests on this? Seems like it may not be doable? Does GitHub provide any service you're aware of that would make testing possible?

clayreimann:
@j-rewerts I think to add tests for this we'd have to create a test user, set them up with OTP, and then commit the secret/salt/QR code/whatever so that we could use something like [this library][opt-lib] to generate the OTP during tests. (That library is not a specific recommendation, just doing a quick search to see if there is likely to be a library that will work)

[otp-lib]: https://www.npmjs.com/package/otplib

j-rewerts:
It seems like it could be possible. I'll sit on this for a bit and try to get something workable together.

PR #461, opened 2017-07-11
Add add label to an issue

asmsuechan:
I could not find these methods so I created. If anyone has any information about them, please let me know.

### DIFF
--- a/lib/Issue.js
+++ b/lib/Issue.js
@@ -258,6 +258,18 @@ class Issue extends Requestable {
    addLabel(issue, label, cb) {
       return this._request('POST', `/repos/${this.__repository}/issues/${issue}/labels`, label, cb);
    }
+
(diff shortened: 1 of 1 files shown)


asmsuechan:
Sure

j-rewerts:
@clayreimann This API is marked as preview by GitHub. What is the policy on using preview APIs?

clayreimann:
@j-rewerts it's unclear. I guess we could mark the API as provisional in this library and let clients assume the risks

j-rewerts:
Yeah that's a good idea. I'll take this code and write tests for them when I have a second. I've added an issue [here](https://github.com/github-tools/github/issues/552) for going through the code base and adding warnings like this.

clayreimann:
Sounds good

PR #462, opened 2017-07-16
Add ability to add collaborator

mtscout6:


.travis.yml
```diff
@@ -1,9 +1,8 @@
 sudo: false
 language: node_js
 node_js:
+  - '8'
```
MarkHerhold: FYI, I opened PR #468 to to test against Node.js v8, which can be closed if this is merged.

j-rewerts:
It seems like [this](https://developer.github.com/enterprise/2.10/v3/repos/collaborators/#add-user-as-a-collaborator) API was considered preview? [Newer docs](https://developer.github.com/v3/repos/collaborators/#add-user-as-a-collaborator) have the URL being `/repos/:owner/:repo/collaborators/:username`, which is the same as the one that was used, so that's nice!

This PR would be easier to review if it didn't also switch the CI and add async await. @clayreimann Is there any plans to change which versions of Node this library builds for? I'm thinking we could merge in the addCollaborator stuff and leave async await for another day.

adjetey0:
This branch has conflicts that must be resolved
Changes can be cleanly merged.

PR #464, opened 2017-07-27
Enable to pass an object to createBlob

### LINKED ISSUE #463: createBlob automatically encodes string content to UTF8
I have web application completely in UTF-8. I define `<meta charset='utf-8' />` and all files are saved in UTF-8 encoding. When I want to create a blob on github, the library automatically encodes the string to UTF8 representation. So it is encoded twice. In my case it breaks everything. For example it transforms `°` to `Â°`.

Why is this doing? I am not expert in this area but I assume that if the HTML page defines charset in meta element, web browser automatically converts strings when it sends some HTTP requests and also set up properly the HTTP headers such as `Content-Type`.

I think the automatic encoding should be removed or there should be option to turn it off? Wdyt?

dudaerich:
createBlob automatically encodes string parameter
by Utf8 library. In some cases the string value
can be already encoded in Utf8 so the encoding
should be skipped.

This commit enables to pass object containing
the content and encoding items. This object is
directly sent to Github. Users have better control
about what is actually sending.

closes #463

### DIFF
--- a/lib/Repository.js
+++ b/lib/Repository.js
@@ -245,7 +245,7 @@ class Repository extends Requestable {
    /**
     * Create a blob
     * @see https://developer.github.com/v3/git/blobs/#create-a-blob
-    * @param {(string|Buffer|Blob)} content - the content to add to the repository
+    * @param {(string|Buffer|Blob|Object)} content - the content to add to the repository
     * @param {Requestable.callback} cb - will receive the details of the created blob
     * @return {Promise} - the promise for the http request
     */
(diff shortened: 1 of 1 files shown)


j-rewerts:
This definitely seems like an important issue. @dudaerich Would you be able to write a test that demonstrates the issue you've run into?

dudaerich:
@j-rewerts sure. You can find it in [this commit](https://github.com/dudaerich/github/commit/6ac67648e6b2bd6555306ca899022c61edb35a18).

j-rewerts:
Could you put that on your issue-463 branch?

dudaerich:
Done. I updated the test to use the object type. Otherwise it would still fail.

To be honest, I am not very convinced that my fix is right. Its advantage is that is backward compatible. On the other hand it is quite unexpected that [this test](https://github.com/dudaerich/github/commit/6ac67648e6b2bd6555306ca899022c61edb35a18) fails. IMO the library shouldn't encode strings to UTF-8. On most platforms UTF-8 is default encoding so it doesn't make sense to encode already encoded strings. If someone uses more exotic encoding, they can encode it to UTF-8 before they pass it to the library. Wdyt?

PR #465, opened 2017-07-27
Added prepare script to package.json

dudaerich:
Since npm@5.0.0 the prepare script is called
when the library is installed from github url.
Aim of the script is to build distribution
files.

See https://github.com/npm/npm/issues/3055

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
   "scripts": {
     "clean": "gulp clean",
     "build": "gulp build",
+    "prepare": "gulp build",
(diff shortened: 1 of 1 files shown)


j-rewerts:
[Here](https://docs.npmjs.com/misc/scripts) is NPM's docs on it. This seems to work for me. It automatically builds the project when I run `npm install`, which is convenient. I'm still not 100% on how this impacts our build process though, since we support Node 4, which doesn't run this. 

I think we should probably wait until we get our tests passing, then rerun our CI for this PR.

PR #485, opened 2017-10-12
Add options to the listReleases function

alexcanessa:
The request to get the list of releases should have the options object in order to pass pagination options.

@see https://developer.github.com/v3/#pagination

### DIFF
--- a/lib/Repository.js
+++ b/lib/Repository.js
@@ -793,11 +793,12 @@ class Repository extends Requestable {
    /**
     * Get information about all releases
     * @see https://developer.github.com/v3/repos/releases/#list-releases-for-a-repository
+    * @param {Object} options - pagination for the list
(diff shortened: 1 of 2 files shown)


j-rewerts:
I'm wondering if there's maybe an alternative way to handle paging. This is going to be an issue with basically every get operation this library does. Would it be crazy to think about doing some sort of async iterables approach? The new `for await ... of` seems pretty cool, but wouldn't work for us, since we also allow people to use callbacks.

What if we used generators and for people using callbacks, we just accumulated the result and returned that. For people using Promises, we could yield Promises until the results are done. It looks like `yield` was added in Node 4, but has no IE support.

@clayreimann Which browsers and versions does github-api support?

j-rewerts:
Just adding options to most GETs would definitely work though.

PR #490, opened 2017-11-17
Add 'raw' parameter to Repository#getBlob.

sudiamanj:
Allow user to choose between Base64 encoded and raw blob, similar to how `Repository#getContents` and `Repository#getReadme` work.

Note: `test/markdown.spec.js` was only updated because the test fails as is. It has nothing to do with my change to `Repository#getBlob`.

### DIFF
--- a/lib/Repository.js
+++ b/lib/Repository.js
@@ -151,7 +151,9 @@ class Repository extends Requestable {
     */
    getBlob(sha, raw, cb) {
       if (typeof raw !== 'boolean') {
-         cb = raw;
+         if (typeof cb === 'undefined') {
+            cb = raw;
+         }
          raw = true;
       }
       return this._request('GET', `/repos/${this.__fullname}/git/blobs/${sha}`, null, cb, raw);


sudiamanj:
Alternatively, we could go with a backwards compatible approach by making `raw` the *last* argument with a default value of `true`. The downside is that is inconsistent with the signatures of `getContents` and `getReadme`.

```javascript
   /**
    * Get a blob from the repository
    * @see https://developer.github.com/v3/git/blobs/#get-a-blob
    * @param {string} sha - the sha of the blob to fetch
    * @param {Requestable.callback} cb - will receive the blob from the API
    * @param {boolean} [raw = true] - `true` if the results should be returned raw instead of GitHub's normalized format
    */
   getBlob(sha, cb, raw = true) {
      return this._request('GET', `/repos/${this.__fullname}/git/blobs/${sha}`, null, cb, raw);
   }
```

Using this function, the `Repository should get blob` test passes in its original form (i.e. not using my updated test suite). Let me know if you'd prefer this version and I'll update the PR accordingly.

PR #502, opened 2018-03-31
Implementing methods to retrieve list of followers and following users.

tomasperezv:
## What

Github v3 provides the endpoints `followers` and `following` associated to the user entity. 

[**List followers of a user**](https://developer.github.com/v3/users/followers/#list-followers-of-a-user)
```
GET /users/:username/followers
```

[**List users followed by another user**](https://developer.github.com/v3/users/followers/#check-if-one-user-follows-another)
```
GET /users/:username/following
```

This change extends the User object by adding the methods `listFollowers` and `listFollowing`, in addition I've added basic unit tests for them.

### DIFF
--- a/lib/User.js
+++ b/lib/User.js
@@ -82,6 +82,26 @@ class User extends Requestable {
    }
 
    /**
+    * List followers of a user
+    * @see https://developer.github.com/v3/users/followers/#list-followers-of-a-user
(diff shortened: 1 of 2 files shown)


j-rewerts:
LGTM. @clayreimann Feel free to merge.

PR #503, opened 2018-04-11
added fork to org

benmonro:


j-rewerts:
This seems like a good idea. Prior to merging, we should set up a test org to actually run this test with successfully.

j-rewerts:
I just realized we already have an org for this purpose. @benmonro Could you switch to using github-api-tests rather than test-org?

j-rewerts:
@clayreimann Does the test user mikedeboertest have access to the org github-api-tests?

clayreimann:
@j-rewerts That's a great question, tbh I'm not sure.

j-rewerts:
Any way for us to find out? We could also rebuild our test user/org setup, but it may be easier to hunt down the existing credentials.

clayreimann:
Could log in as the test use and see. https://github.com/github-tools/github/blob/09384e3981de3ef40caddff62fcefccf6aa89523/test/fixtures/user.json#L2-3

j-rewerts:
😆 Not sure how I didn't think of that.

j-rewerts:
Just confirmed he does have access.

PR #516, opened 2018-07-29
Add stream response type option

xgeek-net:
Hi, there, 
There has a bug when I download binary file with github api, and I found responseType needs to be stream but not text, so can you review this and marge?


j-rewerts:
Hey @xgeek-net Could you write a test that shows what you're talking about?

j-rewerts:
Closing for now. Feel free to reopen with tests. If you can't get around to it, I will eventually.

PR #501, opened 2018-03-30
set the expected result of the markdown GFM unit test to the right value

DanielRuf:

PR #545, opened 2019-04-05
Updated markdown GFM test.

j-rewerts:
As an alternative to PR #501. I'm not sure if the response just changed within the past few days.

I think if we merge this in, and #544 , then tests will be passing.

### DIFF
--- a/test/markdown.spec.js
+++ b/test/markdown.spec.js
@@ -37,7 +37,7 @@ describe('Markdown', function() {
       };
       markdown.render(options)
          .then(function({data: html}) {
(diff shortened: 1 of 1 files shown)


j-rewerts:
Also, if @DanielRuf updates their test, feel free to close this PR in lieu of that one.

DanielRuf:
It seems there were also other changes or the API had issues.

https://travis-ci.org/github-tools/github/builds/516385114

j-rewerts:
Yeah that should be covered in #544. If both are merged, I think the test suite passes.

PR #505, opened 2018-04-15
fixed(markdown): render the markdown in browser.

jacksao:
fixes #451
closes #451

PR #548, opened 2019-04-07
Fix browser markdown

### LINKED ISSUE #451: How can I render the markdown in browser,may be a bug
## Firstly I test this code in node
```javascript
const github = require('github-api');

const gh = new github({
    username: "**********",
    password: "**********"
});
const markdown = gh.getMarkdown();
var options = {
    text: '# Hello world!'
};

markdown.render(options).then(function({data:html}) {
    console.log(html)
});
```
So I get flowing string.
```html
<h1>
<a id="user-content-hello-world" class="anchor" href="#hello-world" aria-hidden="true"><span aria-hidden="true" class="octicon octicon-link"></span></a>Hello world!</h1>
```

## But when I run this code in browser
```html
<script src="GitHub.bundle.js"></script>
<script>
    const gh = new GitHub({
        username: "**********",
        password: "**********"
    });
    const markdown = gh.getMarkdown();
    var options = {
        text: '# Hello world!'
    };

    markdown.render(options).then(function({data:html}) {
        console.log(html)
    });
</script>
```
I get a `null`,so please help me.......

j-rewerts:
Fixes #451. Based on PR #505. 

I also wrote a test to verify the conditions mentioned by the deleted user from #505 and @lvxianchao. We were returning null in the browser.

### DIFF
--- a/lib/Markdown.js
+++ b/lib/Markdown.js
@@ -32,7 +32,7 @@ class Markdown extends Requestable {
     * @return {Promise} - the promise for the http request
     */
    render(options, cb) {
-      return this._request('POST', '/markdown', options, cb);
+      return this._request('POST', '/markdown', options, cb, true);
    }
 }
 
--- a/test/dist.spec/index.html
+++ b/test/dist.spec/index.html
@@ -12,7 +12,7 @@
   <script>mocha.setup('bdd')</script>
   <script>
      describe('the dist file', function() {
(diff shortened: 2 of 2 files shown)


clayreimann:
@j-rewerts I just looked and there's one markdown related test failing on this PR

j-rewerts:
@clayreimann Okay I think it should be good.

Issue #553, opened 2019-04-11
Follow user follows yourself (same for unfollow)


CodyGramlich:
https://github.com/github-tools/github/blob/90fb78b2fb4b4e8e607d24a3dacc983caed2437e/lib/User.js#L182-L202

The `username` argument in these functions is not used. Instead, these requests attempt to follow or unfollow the authenticated user. Surprisngly, there are no errors when you attempt to follow yourself and the same status code is returned as if you were to follow someone else. However, if you make a request to follow yourself, you won't actually be listed as one of your followers.

CodyGramlich:
#513 This person ran into this problem.

PR #554, opened 2019-04-11
Fix follow and unfollow user

CodyGramlich:
Fixes issue #553.

### DIFF
--- a/lib/User.js
+++ b/lib/User.js
@@ -150,7 +150,7 @@ class User extends Requestable {
     * @return {Promise} - the promise for the http request
     */
    follow(username, cb) {
-      return this._request('PUT', `/user/following/${this.__user}`, null, cb);
+      return this._request('PUT', `/user/following/${username}`, null, cb);
    }
 
    /**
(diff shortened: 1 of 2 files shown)


j-rewerts:
@CodyGramlich Could you write some tests that verify the original problem and show how your changes resolve them?
Nice catch though. Surprised this wasn't reported before.

CodyGramlich:
@j-rewerts Sure, I'll make a test by requesting to follow a user, and then requesting to get the user's that I'm following. Then I'll check if the user I tried to follow is in that list. There isn't really an easier way to test it because the response of following yourself is the same response as if you successfully followed someone else.

CodyGramlich:
@j-rewerts I added some tests. One of them is to try to follow yourself and show that you're not one of your followers. The previous test does the same thing but for a different user and it shows that you are now following the user.

j-rewerts:
Sorry for the delay @CodyGramlich. Could you resolve the conflicts?

j-rewerts:
Thanks for the extensive tests! I'm trying out a new review process, so bear with me here.

PR #559, opened 2019-05-03
Repository 'listCommits' working when options omitted

### LINKED ISSUE #508: Repository `listCommits` does not call callback if options are omitted
# Expected Behaviour
Calling `listCommits` on a `Repository` should execute the callback even when no options are given. According to [the function's documentation](http://github-tools.github.io/github/docs/3.0.0/Repository.html#listCommits), the `options` are optional (unlike, for example, the `options` of [`listPullRequests`](http://github-tools.github.io/github/docs/3.0.0/Repository.html#listPullRequests)).

# Actual Behaviour
Calling `listCommits` on a `Repository` without giving an options object results in the callback *not* being called, even when the `Promise` is fulfilled.

# Steps to Reproduce the Problem
I have reduced my own code to the following minimal working example:
```html
<!DOCTYPE html>
<html>
    <body>
        <script src="https://unpkg.com/github-api/dist/GitHub.bundle.min.js"></script>
        <script>
            var gh = new GitHub();

            var repo = gh.getRepo("github-tools", "github");
            repo.listCommits(function() {
                alert("Callback was called.");
            });
        </script>
    </body>
</html>
```
I observed that there was no alert.

In order to verify that the issue was indeed with the lack of options and not with my setup, I changed the above code to include an empty options object:
```html
<!DOCTYPE html>
<html>
    <body>
        <script src="https://unpkg.com/github-api/dist/GitHub.bundle.min.js"></script>
        <script>
            var gh = new GitHub();

            var repo = gh.getRepo("github-tools", "github");
            repo.listCommits({}, function() {
                alert("Callback was called.");
            });
        </script>
    </body>
</html>
```
This time around I *did* receive an alert.

# Specifications
* **Version**: 3.0.0
* **Operating system**: Windows 10 (64 bit)
* **Browser**: Google Chrome, version 66.0.3359.139 (Official Build) (64-bit)

WJXHenry:
Fixes #508.

This PR resolves the issue of having to explicitly specify the options parameter. I also changed the existing test to properly validate a 'listCommits' function call without options.

### DIFF
--- a/lib/Repository.js
+++ b/lib/Repository.js
@@ -187,8 +187,10 @@ class Repository extends Requestable {
     * @return {Promise} - the promise for the http request
     */
    listCommits(options, cb) {
-      options = options || {};
-
+      if (typeof options === 'function') {
+         cb = options;
+         options = {};
+      }
       options.since = this._dateToISO(options.since);
       options.until = this._dateToISO(options.until);
 
--- a/lib/Repository.js
+++ b/lib/Repository.js
@@ -187,6 +187,7 @@ class Repository extends Requestable {
     * @return {Promise} - the promise for the http request
     */
    listCommits(options, cb) {
+      options = options || {};
       if (typeof options === 'function') {
          cb = options;
          options = {};
--- a/test/repository.spec.js
+++ b/test/repository.spec.js
@@ -121,7 +121,7 @@ describe('Repository', function() {
       });
 
       it('should list commits with no options', function(done) {
-         remoteRepo.listCommits(null, assertSuccessful(done, function(err, commits) {
+         remoteRepo.listCommits(assertSuccessful(done, function(err, commits) {
             expect(commits).to.be.an.array();
             expect(commits.length).to.be.above(0);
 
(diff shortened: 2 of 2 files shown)


lib/Repository.js
```diff
@@ -187,8 +187,10 @@ class Repository extends Requestable {
     * @return {Promise} - the promise for the http request
     */
    listCommits(options, cb) {
-      options = options || {};
-
+      if (typeof options === 'function') {
```
j-rewerts: It looks like the options object is not considered optional according to the JS docs. I think it's also worth thinking about existing users, so we don't change the interface for them going forwards without a major semantic version increase.
I think if you include both the default value setting of `options = options || {};` as well as the callback code, it should work for both situations. Also update the Jsdoc comments to mark options as optional.

WJXHenry: I added the original functionality back along with the new code. The options parameter is already marked as optional in the [JSDocs](http://github-tools.github.io/github/docs/3.1.0/Repository.html#listCommits).

j-rewerts:
LGTM. Thanks @WJXHenry .

PR #571, opened 2019-06-17
Fix tests

j-rewerts:
It appears GitHub requires accounts emails to be verified, which was causing our test suite to fail. I've added a new user that we have better control over. I also decided to use a personal access token, rather than exposing the password.

I'm hoping to switch our dev workflow to be more secure. In the past, we exposed our test user credentials publicly. A secure workflow could be:
1. External collaborator submits a PR. No tests run.
2. A project maintainer reviews PR for obvious security issues, or attempts to log environment variables.
3. If 2. is good, merge PR into branch new branch off of master. This triggers a TravisCI build.
4. If Travis is good, review just like we did previously.

This is extra work for sure, but it will allow us to have a secure CI user while also accepting community contributions.


j-rewerts:
The PR tests are from earlier, before I removed them. The push tests passed.

CodyGramlich:
@j-rewerts Won't contributors want to be able to run the tests and possibly make changes to the tests?

j-rewerts:
Yeah so this change means contributors will need to request a private access token, or set up their own test account for running the tests with. I'll add something to the readme.

PR #572, opened 2019-06-17
Following users

j-rewerts:
Fixes issue #553.

Continued from #554.


j-rewerts:
@CodyGramlich Thanks. Looks good to me!

Issue #569, opened 2019-06-17
`_requestAllPages` pagination is broken


connorjclark:
https://github.com/github-tools/github/blob/943ac8a7d36dbac6c0848bee9a5ca64dc229f807/lib/Requestable.js#L259

This bit: 
```
typeof options.page !== 'number'
```

can never be true, as the api functions (well, at least the one I need, `listStarredRepos`) don't allow for passing in options. Besides, it doesn't make much sense, I think it should be negated?

I have this line in my build script to bypass this issue:
```
sed -i '' "s/typeof options.page !== 'number'/false/g" dist/scripts/GitHub.bundle.js
```

tamer1an:
Yes it seems i can only get 100 starred repos

tamer1an:
Here is the fix: https://github.com/github-tools/github/pull/575

j-rewerts:
I'll take look at this soon. thanks!

j-rewerts:
This should be resolved in [3.2.3](https://www.npmjs.com/package/github-api). Please try and confirm.

PR #575, opened 2019-07-17
#569 tamer1an fix  for get all pages in _requestAllPages

tamer1an:
Fixing for #569 _requestAllPages

https://github.com/github-tools/github/issues/569

### DIFF
--- a/lib/Requestable.js
+++ b/lib/Requestable.js
@@ -257,8 +257,11 @@ class Requestable {
 
             const nextUrl = getNextPage(response.headers.link);
             if(nextUrl) {
+               if (!options) {
+                  options = {};
+               }
                options.page = parseInt(
-                 nextUrl.match(/(&page=[0-9]*)/g)
+                 nextUrl.match(/(page=[0-9]*)/g)
                    .shift()
                    .split('=')
                    .pop()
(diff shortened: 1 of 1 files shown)


lib/Requestable.js
```diff
@@ -256,9 +256,17 @@ class Requestable {
             results.push(...thisGroup);
 
             const nextUrl = getNextPage(response.headers.link);
-            if (nextUrl && !(options && typeof options.page !== 'number')) {
-               log(`getting next page: ${nextUrl}`);
-               return this._requestAllPages(nextUrl, options, cb, results);
+            if(nextUrl) {
+               options.page = parseInt(
```
j-rewerts: It's possible for options to be undefined. Please handle this prior to setting page.

lib/Requestable.js
```diff
@@ -256,9 +256,17 @@ class Requestable {
...
+                 nextUrl.match(/(&page=[0-9]*)/g)
```
j-rewerts: This only matches against the page having a parameter before it. Something like this would break it:
``` bash
https://api.github.com/organizations/17532995/teams?page=2
```

Is there a reason we need to match against the &?

lib/Requestable.js
tamer1an: ok will do that

lib/Requestable.js
tamer1an: ok will double check

tamer1an:
Congratz with a new version :muscle:

Thanks for help :green_heart: I almost introduced that fix with `&` but was at holidays

PR #583, opened 2019-09-06
Add pagination to search endpoints

alaycock:
Changes:

- Updated the jsdoc to correctly document the `Github.search` function.
- If a `page` option is passed into `_requestAllPages`, it will only fetch that single page, instead of attempting to fetch all of the pages.
- Removed the `results` parameter from `_requestAllPages`, since that is only meant to be used internally anyway. It's trivial to re-introduce, so let me know if this makes sense to do, since this could break some backwards compatibility if folks are using an unsupported argument.
- Added a test for pagination, and a new fixture for that test. Fixed a number of other fixtures that didn't seem to have the correct page number.

### DIFF
--- a/lib/GitHub.js
+++ b/lib/GitHub.js
@@ -91,11 +91,11 @@ class GitHub {
 
    /**
     * Create a new Search wrapper
-    * @param {string} query - the query to search for
+    * @param {Search.Params} searchParameters - the query and other search parameters
     * @return {Search}
     */
-   search(query) {
-      return new Search(query, this.__auth, this.__apiBase);
+   search(searchParameters) {
+      return new Search(searchParameters, this.__auth, this.__apiBase);
    }
 
    /**
--- a/lib/Requestable.js
+++ b/lib/Requestable.js
@@ -235,13 +235,30 @@ class Requestable {
     * @param {string} path - the path to request
     * @param {Object} options - the query parameters to include
     * @param {Requestable.callback} [cb] - the function to receive the data. The returned data will always be an array.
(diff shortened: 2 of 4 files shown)


lib/GitHub.js
```diff
@@ -91,11 +91,11 @@ class GitHub {
 
    /**
     * Create a new Search wrapper
-    * @param {string} query - the query to search for
+    * @param {Search.Params} searchParameters - the query and other search parameters
```
alaycock: Not sure if jsdoc understands how to reach into other files (Search.js). Let me know if this will be a problem.

lib/Requestable.js
```diff
@@ -256,19 +273,19 @@ class Requestable {
             results.push(...thisGroup);
 
             const nextUrl = getNextPage(response.headers.link);
-            if(nextUrl) {
+            if(nextUrl && !manualPagination) {
                if (!options) {
                   options = {};
                }
                options.page = parseInt(
```
alaycock: If someone can clarify this, I would appreciate it. Why does the page number need to be manually extracted and appended to the options? Since the `nextUrl` is being extracted from the headers, wouldn't it have all the options it already needs? Then by calling `this._requestAllPages(nextUrl, options, cb, results);` the options are being re-added to the url, which results in duplicates in the path, eg:

`/search/repositories?q=tetris+language%3Aassembly&sort=stars&order=desc&type=all&per_page=100&page=4&q=tetris+language:assembly&sort=stars&order=desc&type=all&per_page=100&page=4`

Although this might just be a workaround for other endpoints, since I have't done exploration into how other endpoints paginate.

PR #588, opened 2019-09-19
fix(repository): prevents lib from crashing when not providing option…

hazmah0:
…al arguments

Ran into this issue today described in this old PR: https://github.com/github-tools/github/pull/456

This provides the same fix and adds a test case, though I can't run the tests since it asks me to email jaredrewerts@gmail.com to get access.

### DIFF
--- a/test/repository.spec.js
+++ b/test/repository.spec.js
@@ -385,6 +385,17 @@ describe('Repository', function() {
          }));
       });
 
+      it('should successfully write to repo when not providing optional options argument', function(done) {
+         remoteRepo.writeFile('master', fileName, initialText, initialMessage, undefined, assertSuccessful(done, function() {
+            wait()().then(() => remoteRepo.getContents('master', fileName, 'raw',
+               assertSuccessful(done, function(err, fileText) {
+                  expect(fileText).to.be(initialText);
+
+                  done();
+               })));
+         }));
+      });
+
       it('should rename files', function(done) {
          remoteRepo.writeFile('master', fileName, initialText, initialMessage, assertSuccessful(done, function() {
             wait()().then(() => remoteRepo.move('master', fileName, 'new_name', assertSuccessful(done, function() {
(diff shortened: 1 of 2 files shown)


test/repository.spec.js
```diff
@@ -385,6 +385,17 @@ describe('Repository', function() {
          }));
       });
 
+      it('should successfully write to repo when not providing optional options argument', function(done) {
+         remoteRepo.writeFile('master', fileName, initialText, initialMessage, undefined, assertSuccessful(done, function() {
```
j-rewerts: I think the reason people are running into this bug is because they're using Promises. The 2 option-less calls I'm imagining people are making:
1. Using callbacks:
``` javascript
remoteRepo.writeFile('master', fileName, initialText, initialMessage, assertSuccessFunction)
```

2. Using Promises:
``` javascript
remoteRepo.writeFile('master', fileName, initialText, initialMessage).then(res => {})
```

I'd like tests to verify the Promise approach, as we have many testing writing files with no options provided, but with a callback.

j-rewerts:
Apologies for the delay @hazmah0. I've been traveling.

Please make the requested changes. As for the testing account, feel free to email me and I can get you access, or alternatively, feel free to use one of your own GitHub accounts. Just note that running the test suite will create and edit a bunch of things, so it's best to avoid using your primary account.

j-rewerts:
Also, I'm not too sure how Hacktoberfest counts contributions, but if you need to close and reopen this to get it to count towards your 5, feel free to do that.

test/repository.spec.js
hazmah0: True. Thinking back, I stumbled upon this using async/await. If one were to use the callback approach this would work fine since there is a check in the beginning to see if `options` is a function and in that case use that as the callback.

hazmah0:
No worries about the contribution count. I'll send an email later today, thanks!

j-rewerts:
LGTM!

hazmah0:
Just noticed that the CI is failing the lint task due to a missing semicolon at line 395. I can submit a new PR with that fix later today!

PR #638, opened 2021-01-26
corrected link

luisborit:
Lists the people who the authenticated user follows.

### DIFF
--- a/lib/User.js
+++ b/lib/User.js
@@ -92,7 +92,7 @@ class User extends Requestable {
    }
 
    /**
-    * List users followed by another user
+    * Lists the people who the authenticated user follows.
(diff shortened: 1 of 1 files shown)


lib/User.js
```diff
@@ -93,7 +93,7 @@ class User extends Requestable {
 
    /**
     * List users followed by another user
```
j-rewerts, line 95: Could you update this comment to match? Something like:
Lists the people who the authenticated user follows.

j-rewerts (changes requested):

lib/User.js
luisborit: Sure!

belgradio-rs: 😄

j-rewerts (approved):

PR #649, opened 2021-04-12
Update Repository.js to works with png

wdison:
Using the raw parameter within the options of the writeFile method, this tool works with the sending of png images, it has not been tested with other files, but I think this solution will work for others as well.

```
  writePng(fileNamePng:string,base64Png:any,commitMsg:string=''){
    let initialMessage=commitMsg||'Commit de '+base64Png;
    let option = {encode:false,raw:true}
    const promise = this.bagRepo.writeFile('master', fileNamePng, base64Png, initialMessage, option);
    promise.then((valor:any) => {
      // console.log(valor);
    },this.initRepo);
    return promise;
  }
```

### DIFF
--- a/lib/Repository.js
+++ b/lib/Repository.js
@@ -784,9 +784,9 @@ class Repository extends Requestable {
       return this.getSha(branch, filePath)
          .then((response) => {
             commit.sha = response.data.sha;
-            return this._request('PUT', `/repos/${this.__fullname}/contents/${filePath}`, commit, cb);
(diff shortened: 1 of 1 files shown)


wdison:
Dear @j-rewerts and other merge makers, I hope you're all right.  
  
This change is small very simple, but it brings a lot of gain to the api, it has compatibility with any other preexisting call to this method, and it is very useful for other users.  
  
Please consider this for merging.

GETBACKGANG:
Replying to recent gmail I received from you

GETBACKGANG:
Thank you

